PR #170, opened 2025-04-04
fix: re-enable CI, refactor and fixes

appcypher:
- Re-enable and update GitHub workflows to focus only on monocore package
- Remove monofs and monoutils-store from release configuration
- Add network scope option to temporary sandbox creation
- Improve documentation and examples in management modules
- Fix argument ordering in sandbox run command
- Update default config template with example sandbox
- Remove debug print statements from config application


appcypher:
Best to recreate the PR

PR #253, opened 2025-06-01
feat(cli): add --host support

Sun-ZhenXing:
Add `--host` for msb server

Fixes https://github.com/microsandbox/microsandbox/issues/246


microsandbox-cli/lib/args/msbserver.rs
```diff
@@ -17,6 +17,10 @@ pub struct MsbserverArgs {
     #[arg(short = 'k', long = "key")]
     pub key: Option<String>,
 
+    /// Host address to listen on
+    #[arg(long, default_value = "127.0.0.1")]
+    pub host: String,
+
```
appcypher: Nit: can you add `127.0.0.1` as a constant around here? 

https://github.com/microsandbox/microsandbox/blob/016b58d0c8357ae833320db909f54485bbbec0c1/microsandbox-utils/lib/defaults.rs#L72-L73

Sun-ZhenXing: Of course, I will submit it soon.

PR #269, opened 2025-06-05
Use stable std::io::IsTerminal

xosnrdev:
This PR replaces the usage of the `atty` crate with the stable [`std::io::IsTerminal`](https://doc.rust-lang.org/std/io/trait.IsTerminal.html) for checking terminal status as it's no longer maintained as encouraged [here](https://github.com/softprops/atty/blob/master/README.md?plain=1#L3-L7).


appcypher (approved):
LGTM!

PR #270, opened 2025-06-06
Feature Request: Go SDK - Initial Implementation

Keithagy:
Main project issue: https://github.com/microsandbox/microsandbox/issues/258#issuecomment-2944864874

Discussion points (with reference to Python SDK):
- [x] In this initial implementation, I have preferred writing idiomatic Go over following the Python SDK exactingly. For instance, I removed the `timeout` parameter from some of the key public API and instead let the timeout (whether on client-side or server-side) be controlled via configs upon the standard library's HTTP client. Agree?
- [x] In the start and stop methods, should we poll the server metrics for `isRunning` before updating the client-side container state?
- [x] Proposing a lightweight adapter layer for pluggable logging
- [x] Thoughts on using something like server-sent events for real-time execution streaming?


appcypher:
Great work!
> In this initial implementation, I have preferred writing idiomatic Go over following the Python SDK exactingly. For instance, I removed the timeout parameter from some of the key public API and instead let the timeout (whether on client-side or server-side) be controlled via configs upon the standard library's HTTP client. Agree?

So the timeout is not just an http client timeout. It tells the server to shutdown the sandbox after the timeout. It is not fully implemented yet but that's the plan. 

> In the start and stop methods, should we poll the server metrics for isRunning before updating the client-side container state?

For `start`, when you get a response, the sandbox is already started.
For `stop`, we don't wait to actually check if it has died.

> Proposing a lightweight adapter layer for pluggable logging

That would be nice but I don't know what that entails

> Thoughts on using something like server-sent events for real-time execution streaming?

Yep. That's planned.

KunalSin9h:
@Keithagy @appcypher CI is failing due to small issue, `yaml` and `yml` extension mismatch between `policy` file and `ci` config. 

FIX: https://github.com/microsandbox/microsandbox/pull/271

KunalSin9h:
@appcypher need to update this branch, as main is updated

sdk/go/README.md
```diff
@@ -1,68 +1,256 @@
...
 ```
 
-Make sure your module follows the [Go Module naming conventions](https://go.dev/doc/modules/managing-dependencies#naming_module).
+## Configuration
+
+### Environment Variables
+
+- `MSB_API_KEY`: API key for Microsandbox server authentication
+- `MSB_SERVER_URL`: Microsandbox server URL (default: `http://127.0.0.1:5555`)
+- `MSB_NAMESPACE`: Default namespace (default: `"default"`)
```
appcypher: namespace, unlike the api key and url, isn't an env var.

sdk/go/msb.go
```diff
@@ -0,0 +1,260 @@
...
+type microSandbox struct {
+	cfg       config
+	state     atomic.Uint32 // we use a lightweight primitive to prevent racing starts / stops; every other method is safe to route concurrently to the underlying (thread-safe) http client
+	rpcClient rpcClient
+	starter
+	stopper
+	codeRunner
+	commandRunner
+	metricsReader
+}
```
appcypher: This is actually a nice API design but I would like to suggest a different approach. In the future there will be CommandRunners that are not CodeRunners or even totally different kind of Runners.

My proposal is that we have a base struct — the first few fields in microsandbox look like they will make it in. Then we can have distinct NodeSandbox, PythonSandbox types with these runner interfaces.

appcypher: nit: wondering if it is still idiomatic to namespace the methods instead, `sandbox.metrics.{All(), CPU(), ...}`, like in other SDKs but it is not a deal breaker.

appcypher (commented):
@Keithagy Left some comments.

sdk/go/msb.go
Keithagy: I will make changes to hew closely to the other sdks, but I do have a question from the python SDK:
Why is it that `command` gets its own namespace while the method to execute code (`BaseSandbox.run`) hangs directly off the base abstract class, instead of in a separate `code` namespace? reading this without further context, this special treatment when it comes to the nesting of the `run` methods seems inconsistent, or at the very least underdocumented.

also, it seems a bit more verbose to me than necessary that NodeSandbox and PythonSandbox are independent types when they're both just different implementations of `CodeRunner + CommandRunner` (all diferences between them encapsulated within a language enum that we can pass in as a constructor option -- this seems true regardless of the language we might be trying to model in).

Not big issues, of course -- i will make the changes to present a consistent API with the other SDKs.

Keithagy: Here's the concern I was having: implementing the node and python sandboxes as separate types feels a bit odd to me given that they share exactly the same implementation other than the different language enum.

CleanShot 2025-06-09 at 21 58 39@2x

We could easily reuse a `LangSandbox` type without locking the user in to any `MicroSandbox` "master" interface -- i.e., remain free to add and remove other kinds of sandboxes in the future, and still providing `NewNode/PythonSandbox` constructors for the user to call. The main thing is just that here we'd be making a different set of types available in the Go SDK than the other SDKs. 

CleanShot 2025-06-09 at 22 00 26@2x

let me know what you think -- i will pause here for now
(aside: you'll notice that I've namespaced the methods per your nit)

appcypher: > I will make changes to hew closely to the other sdks, but I do have a question from the python SDK: Why is it that `command` gets its own namespace while the method to execute code (`BaseSandbox.run`) hangs directly off the base abstract class, instead of in a separate `code` namespace? reading this without further context, this special treatment when it comes to the nesting of the `run` methods seems inconsistent, or at the very least underdocumented.

This is a valid point. Other sandboxing services use a similar `sb.run` interface for code run and that's why it is that way. But you are right. We could make it more consistent here.

> also, it seems a bit more verbose to me than necessary that NodeSandbox and PythonSandbox are independent types when they're both just different implementations of `CodeRunner + CommandRunner` (all diferences between them encapsulated within a language enum that we can pass in as a constructor option -- this seems true regardless of the language we might be trying to model in).

Let's do it your way.

Keithagy: > Let's do it your way.

ok. I can find time to help align the py, ts and rs SDKs!
Apologies this small change took a while to turn around -- it's been hectic at work.

Keithagy:
> @Keithagy Looks like you are done with this. Is that correct?

indeed @appcypher. pinged you about it on discord some days back!

PR #310, opened 2025-08-02
[wip] feat: add support for github container registry

OAyomide:


microsandbox-core/lib/oci/pull.rs
```diff
@@ -45,6 +45,18 @@ pub trait OciRegistryPull {
         digest: &Digest,
     ) -> MicrosandboxResult<ImageManifest>;
 
+    /// Fetches an image manifest from Github Container Registry, using the image tag.
+    /// This is optional because Ghcr deviates a bit from Docker as it does not return an image index.
+    ///
+    /// Provides the list of layers and configurations for an image.
+    async fn fetch_ghcr_manifest(
```
OAyomide, line 52: created an optional trait method specifically for ghcr because ghcr doesnt return the `ImageIndex` for a container image but returns a valid single manifest. But we need the `ImageIndex` in order to find the corresponding `Platform` image digest, which is then used to fetch the single manifest.

Implication is that we cannot use `fetch_index` for ghcr and we need to just fetch a single manifest (which the trait definition is also different). So thats why this exists. I am looking for ideas on how to probably not do this but I'm not sure (yet).


as additional context: the github container registry is opaque and also hard to get proper images to use. Perhaps there's a way to fetch `ImageIndex` from GH, perhaps we can also specify specific `Platform` manifest from GH by specifying an appropriate header, so that we can keep the implementations between docker and github more closer. I dont know how to do all that yet, I've tried everything I can from the `OCI spec` in order to get github to return `ImageIndex` but it doesnt work.

cc @appcypher  (important context for this PR)

.gitignore
```diff
@@ -11,6 +11,7 @@ private
 *.tmp
 .history
 .DS_Store
+.zed
```
OAyomide, line 14: here because I use zed. Zed is also becoming popular so I thought it could make sense to acknowledge it and ignore in git

microsandbox-core/lib/oci/implementations/ghcr.rs
```diff
@@ -0,0 +1,752 @@
...
+                format!("@{}:{}", digest.algorithm(), digest.digest())
+            }
+        };
+
+        let image_id = db::save_or_update_image(&self.oci_db, &reference, total_size).await?;
+
+        let platform = Platform::default();
+        let index_id = db::save_index(&self.oci_db, image_id, &index, Some(&platform)).await?;
+
+        let manifest = self.fetch_ghcr_manifest(repository, &reference).await?;
```
OAyomide, line 311: `reference` here would be something like `super-linter@sha256:83d590890328d42191c974cb1dceba0d3ad4a994a1e306e73d50d64218bd23ab` in the case of digest part included.

the problem is that based on earlier comment (regarding github registry api opacity), the ghcr API might not be able to fetch the manifest.

perhaps this is an implementation details that should be noted because the implication here is that we'll only reliably be able to fetch manifest for an image if its `tag` only and `reference` is "latest".

microsandbox-core/lib/oci/implementations/ghcr.rs
```diff
@@ -0,0 +1,752 @@
...
+            result?;
+        }
+
+        #[cfg(feature = "cli")]
+        download_layers_sp.finish();
+
+        Ok(())
+    }
+
+    async fn fetch_index(
```
OAyomide, line 407: to satisfy trait implementation. any other way to satisfy trait without doing this (and other similar cases)

appcypher (approved):

PR #327, opened 2025-11-04
chore(docs): simplify readme

toksdotdev:
this pr attempts to simplify the project's readme in a bid to make it easier to grasp the core.

**notable**
- the use case was moved to a separate file (with a brief reference in the main readme.md 
- the architecture diagram was moved to the development.md file
- major demos were placed together right after the project description
- put less emphasis on the pitfalls of existing solutions, and more on microsandbox's offerings

**future improvements**
- the claude code demo GIF has a very low resolution.
- open tracking issue for SDKs, and update readme with it

**preview**
- https://github.com/toksdotdev/microsandbox/tree/toks/readme-cleanup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,13 @@
 </div>
 <br/>
 
-# <img height="20" src="https://octicons-col.vercel.app/package/A770EF">&nbsp;Microsandbox
+## <img height="20" src="https://octicons-col.vercel.app/package/A770EF">&nbsp;Microsandbox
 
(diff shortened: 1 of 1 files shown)


appcypher (approved):
i like the changes. i agree with moving the use cases out. they were supposed to be actual working demos but that never happened. so better to have them as separate md.
i made some changes to the look a bit which is more of a taste thing than anything else but the structure and content is good.

PR #351, opened 2026-01-05
feat(image): handle layers with missing parent directory for files

toksdotdev:
there are some images where a layer might be missing a parent directory for a file in the layer, but in at least one of the layer's parent, the directory was already defined. this doesn't adhere to the OCI spec (Docker images adhere to this). an example image is: `us-docker.pkg.dev/google-samples/containers/gke/hello-app:1.0`

this PR ensures that if a parent directory doesn't exist in the current layer, an attempt is made to search for the recently modified copy of that directory in prior layers. if the directory is found, it's copied over to the extracted directory of the current layer, alongside all of its attributes and permissions.

to achieve this consistently, i opted to introduce a few dedicated type abstractions over a layer and image.

### DIFF
--- a/microsandbox-core/lib/oci/layer/extraction.rs
+++ b/microsandbox-core/lib/oci/layer/extraction.rs
@@ -16,6 +16,7 @@ use tokio_tar::{Archive, Entry};
 use crate::{MicrosandboxError, MicrosandboxResult, oci::LayerDependencies};
 
 /// Helper function to get full mode with file type bits
+#[allow(clippy::unnecessary_cast)] // libc::S_IF* types differ between platforms (u16 on macOS, u32 on Linux)
 fn get_full_mode(entry_type: &tokio_tar::EntryType, permission_bits: u32) -> u32 {
     let file_type_bits = if entry_type.is_file() {
         libc::S_IFREG as u32
--- a/microsandbox-core/lib/management/config.rs
+++ b/microsandbox-core/lib/management/config.rs
@@ -514,8 +514,9 @@ pub async fn apply_image_defaults(
         tracing::info!("applying defaults from image configuration");
 
         // Apply working directory if not set in sandbox
-        if sandbox_config.get_workdir().is_none() && config.config_working_dir.is_some() {
-            let workdir = config.config_working_dir.unwrap();
+        if sandbox_config.get_workdir().is_none()
+            && let Some(workdir) = config.config_working_dir
+        {
             tracing::debug!("using image working directory: {}", workdir);
             let workdir_path = Utf8UnixPathBuf::from(workdir);
             sandbox_config.workdir = Some(workdir_path);


microsandbox-core/lib/oci/image.rs
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /// Returns the parent layers for a given layer.
+    pub(crate) fn get_layer_parent(&self, digest: &Digest) -> LayerDependencies {
+        let parents = self
+            .layers
+            .rsplit(|layer| layer.digest() == digest)
+            .next()
+            .map(|layer| layer.to_vec())
+            .unwrap_or_default();
```
appcypher, line 80: Do you mean to use `split` here? An `rsplit` splits from the right and with the `next`, you are getting the rightmost segment first.

Say we have 4 layers, [A, B, C, D], where A is the bottom layer and D is the top layer.
If we we want to get the layers below layer C, rsplit would give us [D] then [A, B]. Using `split` would give us [A, B] then [D].

Interesting, rsplit and split would give the same result where there is only one segment to be returned like when you are getting the layers below layer D.

microsandbox-core/lib/oci/layer/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+    fn global_layer_ops(&self) -> &dyn GlobalCacheOps {
+        self.global_layer_ops.as_ref()
+    }
+
+    fn digest(&self) -> &Digest {
+        &self.digest
+    }
+
+    async fn extracted(&self) -> MicrosandboxResult<(bool, OwnedMutexGuard<()>)> {
+        let guard = self.lock.clone().lock_owned().await;
```
appcypher, line 123: This lock seems not to serve any use. Is your intent here to prevent repeat extraction of the same layer?

microsandbox-core/lib/oci/global_cache.rs
```diff
@@ -0,0 +1,206 @@
...
+    fn tar_download_dir(&self) -> &PathBuf {
+        &self.tar_download_dir
+    }
+
+    fn extracted_layers_dir(&self) -> &PathBuf {
+        &self.extracted_layers_dir
+    }
+
+    async fn build_layer(&self, digest: &Digest) -> Arc<dyn LayerOps> {
+        Arc::new(Layer::new(Arc::new(self.clone()), digest.clone()))
```
appcypher, line 131: This part further reinforces the point that the mutex in Layer probably does not do what is intended

microsandbox-core/lib/oci/layer/extraction.rs
```diff
@@ -0,0 +1,415 @@
...
+        }
+
+        let (digest, parent_path) = parent_layers.find_dir(ancestor).await?.ok_or_else(|| {
+            anyhow!(
+                "ancestor directory not found in any parent layer: {}",
+                ancestor.display()
+            )
+        })?;
+
+        let dest_dir = extract_dir.join(dir_name.expect("ancestor directory should have a name"));
```
appcypher: There is an edge-case bug here. `let dir_name = ancestor.components().next()` will always return the first path segment of `ancestors` across iterations and that's what's being used as the destination dir `let dest_dir = extract_dir.join(dir_name...)`.

So if you have `/a/b/c/d/example.txt`:
- 1st iteration: ancestor = `/a`, dir_name = `a`, dest_dir = `extract_dir/a`
- 2nd iteration: ancestor = `/a/b`, dir_name = `a`, dest_dir = `extract_dir/a` (overwrites previous)
- 3rd iteration: ancestor = `/a/b/c`, dir_name = `a`, dest_dir = `extract_dir/a` (overwrites previous)
- ...


Also an opportunity to have a test case for layers with missing deeply-nested parents.

microsandbox-core/lib/oci/layer/mod.rs
toksdotdev: this was intentional. i didn't want extracted checks to happen during an in-progress extraction.

microsandbox-core/lib/oci/global_cache.rs
toksdotdev: uhm, i'm not sure i understand. can you explain? whenever `Layer` is cloned, an `Arc<Mutex>` is cloned as well..

microsandbox-core/lib/oci/layer/mod.rs
appcypher: On a second look, I think I misinterpreted your intent here. You are right. Resolve this

microsandbox-core/lib/oci/global_cache.rs
appcypher: I understand what it does now. Ignore this one.

microsandbox-core/lib/oci/image.rs
toksdotdev: good catch. i've pushed a fix.

microsandbox-core/lib/oci/layer/extraction.rs
toksdotdev: applied a fix to this as well, and also added more tests to verify this behaviour.

appcypher (approved):
LGTM

PR #361, opened 2026-01-12
feat: optimize health check

### LINKED ISSUE #360: optimize health check
The issue: the code treats any HTTP response (including 502) as “connected” and then proceeds to send requests.
Fix: treat only 2xx as a successful connection, or check the actual RPC endpoint instead.

fatelei:
fix #360

### DIFF
--- a/microsandbox-server/lib/payload.rs
+++ b/microsandbox-server/lib/payload.rs
@@ -12,7 +12,7 @@
 //! - Success message formatting for sandbox operations
 //! - Detailed error information handling
 
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value;
 
 //--------------------------------------------------------------------------------------------------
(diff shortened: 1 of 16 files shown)


microsandbox-server/lib/handler.rs
```diff
@@ -270,25 +272,39 @@ pub async fn forward_rpc_to_portal(
 
     // Keep trying to connect until we succeed or hit max retries
     while retry_count < MAX_RETRIES {
-        // Check if portal is available with a HEAD request
+        // Check if portal is available and ready using the health check endpoint
         match client
-            .head(&portal_url)
+            .get(&portal_health_url)
```
toksdotdev: let's use a head here. we only care about the status code.

microsandbox-portal/lib/handler.rs
```diff
@@ -21,6 +21,18 @@ use crate::portal::repl::{Language, start_engines};
 // Functions
 //--------------------------------------------------------------------------------------------------
 
+/// Health check endpoint to verify portal readiness
+pub async fn health_check_handler(
+    State(state): State<SharedState>,
+) -> Result<impl IntoResponse, PortalError> {
+    let ready = *state.ready.lock().await;
```
toksdotdev: let's switch to a `.try_lock()`. it's a health check endpoint, so no need holding unto the lock

toksdotdev (changes requested):
appreciate the contribution. 🙏🏽 asides the comments i raised, lgtm!

microsandbox-portal/lib/handler.rs
toksdotdev: looked at it again, and realised we could have used an `AtomicBool` would the onset.

PR #375, opened 2026-02-15
Add Python SDK CI checks on self-hosted runner

anosvoldi883:
## What this PR does

Adds a GitHub Actions workflow to validate the Python SDK using a self-hosted Linux runner with KVM enabled.

The workflow:
- Installs the Python SDK from source
- Verifies imports and basic API surface
- Runs minimal integration smoke checks against a running Microsandbox server

## Why this is needed

GitHub-hosted runners do not support KVM or Apple Silicon, which prevents running the Microsandbox server in CI.
Using a self-hosted runner allows us to start the server and run real SDK integration checks.

This provides early detection of:
- Packaging or install regressions
- Import/API breakage
- Basic runtime compatibility issues

## Scope / Notes

- This PR focuses **only on the Python SDK** as an initial step
- The workflow is structured so other SDKs (Node, Go, Rust) can be added incrementally
- No changes to SDK runtime behavior

## Follow-ups

- Extend CI coverage to additional SDKs
- Add deeper integration tests as the workflows mature

### DIFF
--- a/.github/workflows/sdk-python-integration.yml
+++ b/.github/workflows/sdk-python-integration.yml
@@ -1,0 +1,46 @@
+name: Python SDK Integration Tests
+
+on:
+  pull_request:
+    paths:
+      - "sdk/python/**"
+      - ".github/workflows/sdk-python-integration.yml"
+
+jobs:
+  python-sdk:
+    name: Python SDK Smoke Tests
+    runs-on: self-hosted-Linux-x86
+
+    defaults:
+      run:
+        working-directory: sdk/python
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/sdk-python-integration.yml
```diff
@@ -0,0 +1,46 @@
...
+on:
+  pull_request:
+    paths:
+      - "sdk/python/**"
+      - ".github/workflows/sdk-python-integration.yml"
+
+jobs:
+  python-sdk:
+    name: Python SDK Smoke Tests
+    runs-on: self-hosted-Linux-x86
```
toksdotdev: nit: 
```suggestion
    runs-on: self-hosted-linux-x64
```

appcypher:
Hey, this CI is based on outdated paradigm. The new re-implementation of [microsandbox](#455) no longer needs a server. 

Python SDK for new implementation is in our short term plans.

PR #469, opened 2026-04-01
chore(deps): update crate deps + reduce oci client verbosity

toksdotdev:
this pr does the following:
- update relevant crate deps
- pins the max log level for `oci-client` to `INFO` until they stop logging the auth token (tracked in https://github.com/oras-project/rust-oci-client/issues/254).


appcypher (approved):
LGTM

PR #474, opened 2026-04-01
docs: add recipe for pulling images from local docker registry

### LINKED ISSUE #315: How to use locally built docker images
Hi there!

Thanks for this amazing project!
I stumbled across it and tried to fiddle around with it but I do not find some information in the code and/or the documentation about how I can use prebuilt docker images as base image for a sandbox.

Currently the workaround is to create a sandbox with ubuntu and have two scripts, one for setup and one for run. the setup then installs a bunch of heavy dependencies.

It would be very cool to have the possibility to run a pre-built docker image locally instead of only be able to pull from the online registry.

Can I do that somehow?

toksdotdev:
## Summary
- Adds a new **Recipe** top-level tab to the docs navigation
- First recipe: **Using Local Docker Images** — step-by-step guide for bridging locally built Docker images to microsandbox via a local registry

Fixes #315

## Test plan
- [x] Verify Mintlify renders the new page correctly
- [x] Confirm navigation shows the Cookbook tab

### DIFF
--- a/docs/recipes/local-images.mdx
+++ b/docs/recipes/local-images.mdx
@@ -1,0 +1,71 @@
+---
+title: Using Local Docker Images
+description: Use locally built Docker images with microsandbox via a local registry
+icon: "docker"
+---
+
+microsandbox pulls images from OCI-compatible registries. If you build an image locally with `docker build`, microsandbox can't access it directly from Docker's local store. The workaround is to run a local registry and push your image there.
+
+## Step 1: Start a local registry
+
+Run a local OCI registry on port 5050:
+
+```bash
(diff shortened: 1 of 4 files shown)


docs/recipes/local-images.mdx
```diff
@@ -0,0 +1,62 @@
+---
+title: Using Local Docker Images
```
mahmudsudo, line 2: isnt it better to word this as `` using locally registered images with microsandbox ``?

toksdotdev: i'll lean towards using terms people are already familiar with. in this case, docker.

PR #482, opened 2026-04-02
docs: replace broken DEVELOPMENT.md references with `just setup` workflow

### LINKED ISSUE #477: Broken link to DEVELOPMENT.md in README.md and CONTRIBUTING.md
`DEVELOPMENT.md` does not exist in the repository. The following links are broken:

https://github.com/superradcompany/microsandbox/blob/5a7a11a92846e6f6fab7029dc17d6e4473e97ef4/CONTRIBUTING.md?plain=1#L70

https://github.com/superradcompany/microsandbox/blob/5a7a11a92846e6f6fab7029dc17d6e4473e97ef4/README.md?plain=1#L380

Please either add the missing file or remove/update these links.

dijdzv:
## Summary

DEVELOPMENT.md was removed in #455 but links in README.md and CONTRIBUTING.md remained, leaving new contributors with broken setup instructions.

This PR replaces those broken references with a working `just setup` recipe and updates the contributing guide accordingly.

### Changes

- **justfile**: Add `just setup` recipe that automates the full dev environment setup (system deps, Rust, submodules, build, install, pre-commit hooks)
- **README.md**: Replace broken DEVELOPMENT.md link with a single reference to CONTRIBUTING.md
- **CONTRIBUTING.md**: Inline setup instructions using `just setup`, fix test command (`cargo test`), fix good-first-issues link pointing to old repo (`monoceros-alpha/monocore`), remove unused link definitions

### Notes

- Linux (`just setup`) has been tested end-to-end on Ubuntu 24.04
- macOS (`just setup`) is **untested** — review and feedback welcome

Closes #477

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2468,7 +2468,7 @@ dependencies = [
 
 [[package]]
 name = "microsandbox"
-version = "0.3.9"
+version = "0.3.10"
 dependencies = [
  "bytes",
  "chrono",
@@ -2505,7 +2505,7 @@ dependencies = [
 
 [[package]]
 name = "microsandbox-cli"
-version = "0.3.9"
+version = "0.3.10"
 dependencies = [
  "anyhow",
  "chrono",
@@ -2530,14 +2530,14 @@ dependencies = [
 
 [[package]]
 name = "microsandbox-db"
-version = "0.3.9"
+version = "0.3.10"
 dependencies = [
  "sea-orm",
 ]
 
 [[package]]
 name = "microsandbox-filesystem"
-version = "0.3.9"
+version = "0.3.10"
 dependencies = [
  "libc",
  "microsandbox-utils",
@@ -2550,7 +2550,7 @@ dependencies = [
 
 [[package]]
 name = "microsandbox-image"
-version = "0.3.9"
+version = "0.3.10"
 dependencies = [
  "astral-tokio-tar",
  "async-compression",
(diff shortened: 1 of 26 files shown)


justfile
toksdotdev, line 1: hesitant with automatically passing the `-y` flag to the apt install and rust install commands. developers should explicitly approve packages being installed. we can allow `-y` flag in `just install`

justfile
```diff
@@ -2,6 +2,77 @@
...
+    echo "==> Checking prerequisites..."
+    command -v docker &>/dev/null || { echo "error: Docker is required on macOS. Install Docker Desktop."; exit 1; }
+    xcode-select -p &>/dev/null || { echo "==> Installing Xcode Command Line Tools..."; xcode-select --install; }
+
+    echo "==> Checking Rust toolchain..."
+    if ! command -v rustup &>/dev/null; then
+        echo "    Rust not found. Installing via rustup..."
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+        source "$HOME/.cargo/env"
+    fi
```
toksdotdev, line 59: nit: this can be a subcommand: say, `_install_dev_deps`. and we can then have a single `setup` command given they're identical. pre commit can also be in that step.

justfile
dijdzv: Removed `-y` from both `apt install` and `rustup`.  6223d28

One thought: we could also accept an optional argument so that `just setup -y` passes it through for users who want non-interactive installs. Would that be worth adding, or keep it simple for now?

justfile
dijdzv: Extracted OS-specific steps into `_install-dev-deps` (split by `[linux]`/`[macos]`) with a single `setup` recipe that calls it as a dependency. 6223d28

PR #507, opened 2026-04-05
feat(agentd): introduce AgentdConfig to read env vars once at startup

dijdzv:
## Summary

`test_pty_reader_drains_ready_fd` fails intermittently (~10%) under `cargo test` in `crates/agentd/`. `test_request_user_overrides_env_default` sets `MSB_USER=0:0` via `set_var`, which leaks into the parallel PTY test's `fork()` — the child picks it up, attempts a privilege switch, and `_exit(1)`s.

The root cause is that `resolve_requested_user` reads `ENV_USER` as a fallback when `ExecRequest.user` is `None`. But the SDK already resolves `config.user` into `ExecRequest.user` in `build_exec_request()`, so this env var fallback never fires in normal SDK/CLI usage — it's effectively dead code.

## Changes

- **session.rs**: Remove `ENV_USER` fallback from `resolve_requested_user`, relying solely on `ExecRequest.user`
- **session.rs**: Remove `test_request_user_overrides_env_default` (the code path it tested was removed above)
- **session.rs**: Reduce PTY test `sleep` (1s → 0.1s) — original delay was unnecessary for verifying output ordering

## Verification

Before: 3 failures in 30 runs
After: 0 failures in 80 runs

### DIFF
--- a/crates/agentd/lib/config.rs
+++ b/crates/agentd/lib/config.rs
@@ -1,0 +1,834 @@
+//! Agentd configuration, read once from environment variables at startup.
+//!
+//! [`AgentdConfig::from_env`] reads all `MSB_*` environment variables and
+//! parses them into their respective types in a single step. Downstream
+//! functions receive the config by reference, avoiding repeated env var reads
+//! and repeated parsing.
+
+use std::net::{Ipv4Addr, Ipv6Addr};
+
+use microsandbox_protocol::{
+    ENV_BLOCK_ROOT, ENV_DIR_MOUNTS, ENV_FILE_MOUNTS, ENV_HOSTNAME, ENV_NET, ENV_NET_IPV4,
+    ENV_NET_IPV6, ENV_TMPFS, ENV_USER,
+};
+
+use crate::error::{AgentdError, AgentdResult};
+
+//--------------------------------------------------------------------------------------------------
+// Types
+//--------------------------------------------------------------------------------------------------
+
+/// Parsed configuration for agentd.
+///
+/// All `MSB_*` environment variables are read and parsed into their respective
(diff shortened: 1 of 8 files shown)


crates/agentd/lib/config.rs
```diff
@@ -0,0 +1,76 @@
...
+            block_root: read_env(ENV_BLOCK_ROOT),
+            mounts: read_env(ENV_MOUNTS),
+            tmpfs: read_env(ENV_TMPFS),
+            hostname: read_env(ENV_HOSTNAME),
+            net: read_env(ENV_NET),
+            net_ipv4: read_env(ENV_NET_IPV4),
+            net_ipv6: read_env(ENV_NET_IPV6),
+            user: read_env(ENV_USER),
+        }
+    }
```
appcypher: Just one more nit 😅.

It would be nice if all the env parsing operations are moved into this file as well.

So we parse all the env vars once in `from_env` and fields in `AgentdConfig` can be the actual types. Something like 

```rs
pub(crate) struct AgentdConfig<'a> { 
    block_root: Option<BlockRootSpec<'a>>, 
    mounts: Option<VolumeMountSpec<'a>>, 
    ... 
}`

dijdzv: Moved all env parsing into `config.rs`. Spec types own `String` instead of borrowing — keeps the API simple with no lifetimes and a single-step `from_env()`. a907292

appcypher:
I need to add a couple patches to this and it should be okay to merge

appcypher (approved):

appcypher:
Sorry for taking so long. Last week was our launch week. 
Thanks again for the contribution!

PR #517, opened 2026-04-08
test(sdk): add Node.js SDK smoke tests and CI workflow

Eronmmer:
Adds smoke tests for the Node.js SDK and a CI workflow to run them on the self-hosted runner.

### Tests

Covers the core SDK surface:
- Create a sandbox (`Sandbox.create`)
- Execute commands (`exec`, `shell`)
- Filesystem operations (`fs.write`, `fs.readString`, `fs.exists`, `fs.stat`)
- Metrics (`sandbox.metrics`)
- List sandboxes (`Sandbox.list`)
- Stop (`sandbox.stopAndWait`)
- Setup check (`isInstalled`)

Uses vitest. Tests run sequentially since they share a single sandbox lifecycle (create → use → stop).

### CI workflow

New `sdk-test.yml` workflow that:
- Runs on `self-hosted-Linux-x86` (needs KVM for sandbox VMs)
- Builds agentd, libkrunfw, msb, and the NAPI binary from source
- Triggers on changes to `sdk/`, `crates/`, or the workflow file itself

### Other changes

- Added `vitest` and `@types/node` as dev dependencies
- Added `npm test` script to `package.json`


sdk/node-ts/tests/smoke.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+const SANDBOX_NAME = "sdk-smoke-test";
+
+describe("Node.js SDK Smoke Tests", () => {
+  let sandbox: Awaited<ReturnType<typeof Sandbox.create>>;
+
+  afterAll(async () => {
+    try {
+      await Sandbox.remove(SANDBOX_NAME);
+    } catch {}
```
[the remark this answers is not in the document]
Eronmmer: resolved

sdk/node-ts/tests/smoke.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    sandbox = await Sandbox.create({
+      name: SANDBOX_NAME,
+      image: "alpine",
+      cpus: 1,
+      memoryMib: 512,
+      replace: true,
+    });
+
+    expect(await sandbox.name).toBe(SANDBOX_NAME);
+  });
```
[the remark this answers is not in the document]
Eronmmer: not sure this is necessary. resolving for now

sdk/node-ts/tests/smoke.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    expect(metrics.cpuPercent).toBeGreaterThanOrEqual(0);
+    expect(metrics.memoryBytes).toBeGreaterThan(0);
+    expect(metrics.memoryLimitBytes).toBe(512 * 1024 * 1024);
+    expect(metrics.diskReadBytes).toBeGreaterThan(0);
+    expect(metrics.diskWriteBytes).toBeGreaterThan(0);
+    expect(metrics.netRxBytes).toBeGreaterThanOrEqual(0);
+    expect(metrics.netTxBytes).toBeGreaterThanOrEqual(0);
+    expect(metrics.uptimeMs).toBeGreaterThan(0);
+    expect(metrics.timestampMs).toBeGreaterThan(0);
+  });
```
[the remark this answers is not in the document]
Eronmmer: not sure this is necessary but leaving open for now

sdk/node-ts/tests/smoke.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  it("should create a sandbox", async () => {
+    sandbox = await Sandbox.create({
+      name: SANDBOX_NAME,
+      image: "alpine",
+      cpus: 1,
+      memoryMib: 512,
+      replace: true,
+    });
```
toksdotdev: the sandbox should be created in a `.beforeAll(..)`. you can still keep the sandbox name assertion in a dedicated test.

sdk/node-ts/vitest.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    testTimeout: 120_000,
+    hookTimeout: 60_000,
+    fileParallelism: false,
```
toksdotdev: this should be true. tests suites should run in paralle.

what we don't want is tests within the sandbox smoke test suite to run in parallel, and that's controlled by [`concurrent`](https://vitest.dev/api/describe.html#describe-concurrent), and by default, it's disabled.

sdk/node-ts/package.json
```diff
@@ -20,15 +20,18 @@
...
     "build": "napi build --release --platform --js index.cjs --dts index.d.cts",
     "build:debug": "napi build --platform --js index.cjs --dts index.d.cts",
+    "test": "vitest run",
     "postinstall": "node postinstall.js"
   },
   "devDependencies": {
-    "@napi-rs/cli": "^3"
+    "@napi-rs/cli": "^3",
+    "@types/node": "^25.5.2",
+    "vitest": "^2"
```
toksdotdev: please use vitest `^4.1`

toksdotdev (changes requested):

sdk/node-ts/tests/smoke.test.ts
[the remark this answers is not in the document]
toksdotdev: valid point. there's the possibility this could be flaky unless a disk I/O operation is guaranteed to always to be done in the background. 

we can keep as is for now, and always update if we realise it's flaky.

.github/workflows/sdk-test.yml
toksdotdev, line 1: the self-hosted-Linux-x86 machine isn't as beefy as the rest of the other machines, and so a full build would be slower. a thought ... re-use msb artifacts from the other CI run (on same architecture), and only do the npm install, build and run smoke test on this machine.

quick search, and found this link: https://docs.github.com/en/actions/tutorials/store-and-share-data

sdk/node-ts/package.json
Eronmmer: updated

sdk/node-ts/vitest.config.ts
Eronmmer: 👍🏼

sdk/node-ts/tests/smoke.test.ts
Eronmmer: addressed

.github/workflows/sdk-test.yml
Eronmmer: i split the workflow into two jobs. `build` on `ubuntu-latest` uploads the artifacts, `node-sdk` on the self-hosted runner just downloads them and runs npm install, build, and test. the new job still seems stuck in queued. is the self-hosted runner online?

sdk/node-ts/tests/smoke.test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+	afterAll(async () => {
+		if (sandbox) {
+			try {
+				await sandbox.stopAndWait();
+			} catch {}
+		}
+		try {
+			await Sandbox.remove(SANDBOX_NAME);
+		} catch {}
```
toksdotdev: let's avoid wrapping this in a try catch. if something goes wrong here, the error should be bubbled up.

.github/workflows/sdk-test.yml
```diff
@@ -0,0 +1,130 @@
...
+    needs: build
+    runs-on: self-hosted-ubuntu-2404-x64
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2
```
toksdotdev: this is no longer needed given the rust project isn't built in this step

.github/workflows/sdk-test.yml
toksdotdev: a simpler approach could be to have the node job within the `check.yml` file. that'd avoid rebuilding, and trying to figure out how to share artifacts across different workflow files.

Eronmmer:
rebased on main and squashed into a single commit to resolve conflicts. All review feedback have been applied, but the `self-hosted-Linux-x86` runner hasn't picked up any job yet @toksdotdev

toksdotdev:
the runner name is `self-hosted-ubuntu-2404-x64`. fixed it in a prior commit, but seems it got lost during the squash 😅

Eronmmer:
using the correct runner name now & the SDK tests are running but the sandbox VM crashes with `SIGABRT ` before booting. wondering if the host logs gives us anything useful to debug with @toksdotdev

toksdotdev (approved):

toksdotdev (approved):

toksdotdev:
thanks for getting this over the finish line. a rust equivalent would also be nice :)

PR #519, opened 2026-04-08
fix(db): make SQLite migrations idempotent and enable WAL mode

appcypher:
## Summary
- Make storage migration 3 idempotent by adding `IF NOT EXISTS` to both snapshot index creations, preventing failures when indexes already exist from a prior partial migration run
- Enable WAL journal mode, 5-second busy timeout, and foreign key enforcement on all SQLite connections (global DB pool in `db/mod.rs` and sandbox runtime in `vm.rs`)
- This fixes `SQLITE_BUSY` errors that occurred when multiple processes (CLI commands + long-lived sandbox runtimes) accessed the shared `~/.microsandbox/db/msb.db` concurrently
- Add a regression test that simulates a half-applied migration 3 (tables and first index exist but migration not recorded) and verifies recovery

## Test Plan
- [x] Run `cargo test -p microsandbox` to verify the new `test_connect_and_migrate_recovers_from_partial_storage_migration` test passes
- [x] Run `cargo test` to ensure no regressions across the workspace
- [x] Run `cargo build` to verify successful compilation
- [x] On a Linux host, delete `~/.microsandbox` and run `msb run bash` to verify fresh DB creation with WAL mode works end-to-end
- [x] Verify concurrent access: run `msb run bash` in one terminal while running `msb sandbox list` in another to confirm no SQLITE_BUSY errors

### DIFF
--- a/crates/microsandbox/lib/db/mod.rs
+++ b/crates/microsandbox/lib/db/mod.rs
@@ -6,7 +6,10 @@
 
 pub use microsandbox_db::entity;
 
-use std::path::{Path, PathBuf};
+use std::{
+    path::{Path, PathBuf},
+    time::Duration,
+};
 
 use microsandbox_migration::{Migrator, MigratorTrait};
 use sea_orm::{ConnectOptions, Database, DatabaseConnection};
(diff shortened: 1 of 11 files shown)


crates/microsandbox/lib/db/mod.rs
```diff
@@ -123,21 +126,36 @@ async fn connect_and_migrate(
...
     Ok(conn)
 }
 
+fn configure_sqlite_connect_options(opts: &mut ConnectOptions, max_connections: u32) {
+    // The global microsandbox DB is shared across short-lived CLI processes and
+    // long-lived sandbox runtimes. WAL mode plus a busy timeout makes that
+    // multi-process pattern much more reliable than the default rollback
+    // journal on Linux hosts.
+    opts.max_connections(max_connections)
+        .connect_timeout(Duration::from_secs(30))
```
toksdotdev: we should make this configurable via config.json (maybe the busy timeout as well?)

toksdotdev (approved):

appcypher:

PR #520, opened 2026-04-08
Add --allow-private-ips CLI flag for NetworkPolicy::allow_all

### LINKED ISSUE #518: Expose CLI flag to allow access to host/private IPs (`NetworkPolicy::allow_all`)
### The Problem
Currently, it is impossible to access services running on the host machine (e.g., a local Ollama instance on port 11434) or other private LAN IPs from inside a microsandbox via the CLI. 

Even when configuring the host service to listen on `0.0.0.0` and using the host's actual local IP (e.g., `192.168.x.x`), the connection instantly fails. This appears to be because the CLI enforces `NetworkPolicy::public_only()` by default, which drops all packets destined for private subnets.

While `--no-dns-rebind-protection` allows DNS to *resolve* to local IPs, the underlying packet filter still blocks the traffic. 

### Use Case
Running local AI development environments inside `msb` where the sandbox needs to communicate with large, resource-heavy services running on the host machine (like an LLM inference server). 

Currently, the only local workaround is to use a clunky file-system bypass (using `socat` to pipe traffic through a Unix socket in a shared volume).

### Proposed Solution
Provide a flag in the `msb create` CLI to override the default network policy and allow traffic to private IP addresses. 

For example, an `--allow-private-ips` or `--network-policy allow_all` flag that passes `NetworkPolicy::allow_all()` to the underlying sandbox configuration.

**Expected CLI Usage:**
```bash
msb run --allow-private-ips ubuntu
```

### Environment Context
* **Host Service:** Ollama bound to `0.0.0.0:11434`
* **Command:** `msb exec dev -- curl http://<HOST_LAN_IP>:11434/v1`
* **Result:** `curl: (7) Failed to connect ... after 0 ms: Could not connect to server`

Bahtya:
## Summary

Fixes #518 by adding a `--allow-private-ips` CLI flag to `msb run` and `msb create` that enables network access to host/LAN private IP addresses.

## The Problem

Currently, the CLI enforces `NetworkPolicy::public_only()` by default, which blocks all traffic to private subnets. This makes it impossible to access services on the host machine (e.g. a local Ollama instance on port 11434) from inside a microsandbox.

While `--no-dns-rebind-protection` allows DNS to resolve to local IPs, the underlying packet filter still blocks the traffic.

## Changes

- Add `--allow-private-ips` boolean flag to `SandboxOpts` (shared by `msb run` and `msb create`)
- When set, apply `NetworkPolicy::allow_all()` via the network builder closure in `apply_network_opts`
- Register the flag in `has_creation_flags()` so it's properly tracked

## Usage

```bash
# Access host Ollama from inside sandbox
msb run --allow-private-ips ubuntu -- curl http://192.168.1.100:11434/v1

# Create a sandbox with private IP access
msb create --name dev --allow-private-ips ubuntu
```

## Testing

- `cargo check -p microsandbox-cli --features net` passes
- Flag is correctly threaded through `SandboxOpts` → `apply_sandbox_opts` → `apply_network_opts` → `NetworkBuilder::policy(allow_all())`

Fixes #518


toksdotdev:
thanks for the pr! 

could you adapt the change to reflect the feedbacks highlighted in https://github.com/superradcompany/microsandbox/issues/518#issuecomment-4207584005. can you'd need to also update the [docs](https://docs.microsandbox.dev/cli/sandbox-commands) to reflect new flags.

nramirezuy:
I need this but I would like it to be more explicit about what I allow.

For example:
```
--allow-cidr 0.0.0.0/0       # allows all IPv4
--allow-cidr 192.168.0.0/24  # allows 192.168.0.*
--allow-cidr 192.168.0.36/32 # allows a single host
--allow-ip 192.168.0.36      # shorthand for a single host
```

Is it possible, or is it too much hassle?

Bahtya:
Thanks for the feedback! I've updated the PR to implement the `--network-policy` flag approach as discussed in #518.

Changes:
- Replaced `--allow-private-ips` boolean with `--network-policy <POLICY>` option
- Supported values: `none`, `public-only`, `nonlocal`, `allow-all`
- Added `NetworkPolicy::non_local()` constructor to the network crate (allows public + private/LAN, blocks loopback/link-local/metadata)

Usage example:
```bash
# Access host Ollama from inside sandbox
msb run --network-policy nonlocal ubuntu -- curl http://192.168.1.100:11434/v1

# Completely unrestricted
msb run --network-policy allow-all ubuntu -- bash
```

@nramirezuy — does this address your request for more explicit control over what's allowed?

Bahtya:
Also pushed a docs update — added `--network-policy` to the [sandbox-commands](docs/cli/sandbox-commands.mdx) options table with all four accepted values and their meanings.

appcypher:
> I need this but I would like it to be more explicit about what I allow.
> 
> For example:
> 
> ```
> --allow-cidr 0.0.0.0/0       # allows all IPv4
> --allow-cidr 192.168.0.0/24  # allows 192.168.0.*
> --allow-cidr 192.168.0.36/32 # allows a single host
> --allow-ip 192.168.0.36      # shorthand for a single host
> ```
> 
> Is it possible, or is it too much hassle?

I'd suggest that this level of granular support gets moved into its own PR.

Bahtya:
Agreed — the `--allow-cidr` / `--allow-ip` granularity is a great next step but deserves its own PR. The current `--network-policy` with the four preset values covers the most common use cases and keeps the initial change focused.

I'm happy to follow up with the CIDR-based API in a separate PR after this one lands if there's interest.

nramirezuy:
@Bahtya I went ahead and left a comment on [the discussion](https://github.com/superradcompany/microsandbox/issues/518#issuecomment-4214454447).

`--network-policy` sounds good to me; thinking about it my purposed granularity will raise the question about ports and protocols.

Bahtya:
Hi @toksdotdev and @appcypher — just a gentle ping here. All CI checks are passing and the implementation has been updated to reflect the feedback from #518:

- `--allow-private-ips` has been replaced with `--network-policy <POLICY>`
- Four supported values: `none`, `public-only`, `nonlocal`, `allow-all`
- `NetworkPolicy::non_local()` constructor was added to the network crate
- Docs updated with the new flag and all accepted values

This matches the approach outlined by @appcypher in the issue. Would love to get this reviewed and merged when you have a chance!

toksdotdev (approved):

PR #521, opened 2026-04-08
feat(sdk): add filesystem readStream to TypeScript SDK

toksdotdev:
## Summary
- Expose the Rust SDK's `read_stream` API to TypeScript via NAPI-RS
- Support both `recv()` and `for await...of` using native `#[napi(async_iterator)]` with `AsyncGenerator`
- Remove "coming soon" labels from filesystem streaming docs
- Add Rust and TypeScript examples for streaming file reads

## Test plan
- [x] Run TypeScript example: `cargo run --manifest-path examples/rust/fs-read-stream/Cargo.toml`
- [x] Run Rust example: `cd examples/typescript/fs-read-stream && npm install && npm start`
- [x] Verify `for await (const chunk of stream)` works
- [x] Verify `stream.recv()` loop works
- [x] Verify file size assertion passes (10 MiB)


sdk/node-ts/tsconfig.json
toksdotdev, line 1: added this because `code` couldn't recognise the node in-built types.

appcypher (approved):
Thanks for fixing the docs. LGTM!

PR #524, opened 2026-04-09
fix(agentd): fix flaky PTY drain test

toksdotdev:
## Summary
- Fix `test_pty_reader_drains_ready_fd` intermittent failure where exit code is 1 instead of 0
- Root cause: the test writes 4 KB through a PTY in a tight loop, which can overflow the PTY buffer and cause `printf` to fail, making the shell exit before reaching `exit 0`
- Reduce output from 1024 to 256 iterations (4 KB → 1 KB), which still exercises the drain logic without risking buffer overflow

## Test plan
- [x] Verify `test_pty_reader_drains_ready_fd` passes consistently in CI

### DIFF
--- a/crates/agentd/lib/session.rs
+++ b/crates/agentd/lib/session.rs
@@ -927,7 +927,7 @@ mod tests {
             cmd: "/bin/sh".to_string(),
             args: vec![
                 "-c".to_string(),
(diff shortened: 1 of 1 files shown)


appcypher (approved):
LGTM

PR #526, opened 2026-04-09
fix(image): unblock layer pulls and detect case collisions

appcypher:
## Summary
- Switch the CLI from a current-thread Tokio runtime to a multi-thread runtime so layer download, extract, and index pipelines can actually progress in parallel
- Extract `flock()` helpers into a new `crates/image/lib/lock.rs` module and run lock acquisition via `spawn_blocking`, so a parked lock waiter no longer pins a runtime worker
- Move post-download SHA-256 verification onto `spawn_blocking` as well, keeping multi-hundred-MB hashing off the async worker threads
- Log a warning when a manifest contains duplicate layer digests so the serialized per-layer processing becomes observable
- Probe the extraction destination for case sensitivity and fail fast with a clear error when a layer contains paths that differ only by case on a case-insensitive host filesystem

Motivation: `msb pull ghcr.io/boldsoftware/exeuntu` stalled at "resolving layer". The amd64 manifest lists the same layer digest three times; on the single-threaded runtime the first pipeline took the per-layer `flock()` and moved to async I/O, while the second pipeline blocked the one runtime thread inside a synchronous `flock()` call, preventing the first from ever resuming — a self-deadlock. A stale prior `msb` holding the per-image lock reproduced the same hang on any repeat pull. Even after the hang was gone, `exeuntu` still failed on default macOS because one of its layers contains both `usr/share/man/man2/_Exit.2.gz` and `usr/share/man/man2/_exit.2.gz`, which collapse onto the same path on case-insensitive APFS and surfaced as an opaque `ELOOP` error.

Impact: image pulls with duplicate layer digests no longer deadlock the CLI; hashing and lock waits no longer starve runtime workers; and users hitting the APFS case-collision case now get an actionable error pointing them at a case-sensitive cache volume instead of "Too many levels of symbolic links". Note: this is diagnostic only for the case-collision path — a proper fix (extraction onto a case-sensitive backing store such as an APFS sparse image) is still required to actually pull images like `exeuntu` on default macOS.

## Test Plan
- [x] `cargo build -p microsandbox-image` and `cargo build -p microsandbox` build cleanly
- [x] `cargo test -p microsandbox-image` passes, including the new `test_record_casefold_path_rejects_case_only_collisions` and `test_extract_layer_rejects_case_only_collisions_on_case_insensitive_fs` tests
- [x] `cargo clippy --workspace --all-targets` is clean
- [x] `msb pull ghcr.io/library/alpine` still succeeds end to end (baseline, no duplicate layers, case-sensitive-safe)
- [x] `msb pull ghcr.io/boldsoftware/exeuntu` no longer stalls at "resolving layer"; on macOS it fails fast with the "differ only by case" error message instead of hanging or returning `ELOOP`
- [x] With a stale lock file present at the per-image lock path, a second `msb pull` of the same reference blocks on the flock but does not freeze the runtime — other concurrent layer tasks in the same process continue to make progress
- [x] Trace logs (`RUST_LOG=microsandbox_image=warn`) emit the "manifest contains duplicate layer digest" warning when pulling an image with duplicated layers

### DIFF
--- a/crates/cli/bin/main.rs
+++ b/crates/cli/bin/main.rs
@@ -160,9 +160,10 @@ fn run_async_command(
     command: Commands,
     _log_level: Option<microsandbox::LogLevel>,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    // CLI commands are foreground and short-lived, so a current-thread
-    // runtime avoids worker startup overhead on each invocation.
-    let runtime = tokio::runtime::Builder::new_current_thread()
+    // Pulls, extraction, and indexing fan out across layers, so the CLI
+    // needs worker threads to realize that parallelism without one task
+    // stalling all others.
+    let runtime = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
         .build()?;
 
--- a/crates/image/lib/layer/extraction.rs
+++ b/crates/image/lib/layer/extraction.rs
@@ -5,7 +5,7 @@
 //! platform-aware symlinks, special file handling, and whiteout markers.
 
 use std::{
-    collections::BTreeSet,
+    collections::{BTreeSet, HashMap},
     io::Read,
     path::{Component, Path, PathBuf},
     pin::Pin,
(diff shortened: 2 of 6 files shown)


toksdotdev (approved):

appcypher:

PR #530, opened 2026-04-10
fix(network): update Content-Length after body secret substitution

toksdotdev:
## Summary
- Fix `SecretsHandler::substitute()` to update the `Content-Length` header when body substitution changes the body size
- Only updates when a Content-Length header already exists (no-op for chunked encoding)
- Add tests for different-length substitution, missing Content-Length, and header-only substitution

## Test plan
- [x] `body_injection_updates_content_length` — placeholder shorter than secret, verifies Content-Length is updated
- [x] `body_injection_no_content_length_header` — chunked encoding, verifies no Content-Length is injected
- [x] `header_only_substitution_preserves_content_length` — body unchanged, verifies Content-Length is untouched
- [x] All 11 existing handler tests pass


appcypher (approved):
LGTM

PR #534, opened 2026-04-10
fix(tls): prevent data loss in TLS MITM proxy under parallel connections

dijdzv:
## Summary

Fixes data loss that caused `cannot decrypt peer's message` errors when
clients (e.g. Claude CLI / Node.js) opened multiple simultaneous TLS
connections through the MITM proxy.

Two independent root causes:

- **conn.rs**: `recv_slice()` irreversibly consumes data from the smoltcp
  socket, but `try_send()` silently dropped it when the mpsc channel was
  full. Added `read_buf` field to retain the data and retry on the next poll.

- **proxy.rs**: `read_tls()` may partially consume input when rustls'
  internal deframer buffer is near capacity. The remainder was silently
  lost. Now loops until all bytes are consumed.

Also installs the CA cert as both `.pem` and `.crt` in guest trust
directories for `update-ca-certificates` compatibility (agentd).

## Test plan

- [x] `cargo test -p microsandbox-network` (56 passed)
- [x] `cargo test` in `crates/agentd` (38 passed)
- [x] E2E stress test: `msb create --tls-intercept --secret` + Claude CLI — 99/100 passed (upstream main: 0/100)


toksdotdev (approved):

toksdotdev:
thanks for the fix! much appreciated. 🙏🏽

PR #536, opened 2026-04-11
feat(install): add Homebrew tap support for macOS

### LINKED ISSUE #531: Support Homebrew (brew) installer
Add a Homebrew formula so users can install microsandbox via `brew install microsandbox`.

toksdotdev:
## Summary
- Add an `update-homebrew` job to the release workflow that dispatches a formula update to [`superradcompany/homebrew-tap`](https://github.com/superradcompany/homebrew-tap) on each release
- Update README with Homebrew install instructions alongside the existing install script

The tap repo already has the receiving `update-formula` workflow, the `microsandbox.rb` formula, and the `HOMEBREW_TAP_GITHUB_PAT` secret is configured, so this closes the loop end to end. The dispatch runs after `assemble` so the GitHub release and `checksums.sha256` exist before the tap pulls them.

## Install
```sh
brew install superradcompany/tap/microsandbox
```

Closes #531

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -283,6 +283,23 @@ jobs:
             release-artifacts/*
 
   # ---------------------------------------------------------------------------
+  # Update Homebrew formula in the tap repo
+  # ---------------------------------------------------------------------------
+  update-homebrew:
+    name: Update Homebrew Formula
+    needs: assemble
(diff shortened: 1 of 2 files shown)


toksdotdev:
this is currently blocked until we've stood up an apple developer account

PR #540, opened 2026-04-13
fix(cli): make command optional for `msb exec` in interactive mode

toksdotdev:
## Summary

- `msb exec -t worker` (no `-- <command>`) now falls back to the sandbox's configured shell instead of erroring with "required arguments were not provided"
- Matches the documented behavior in `docs/sandboxes/commands.mdx` and aligns with how `msb run` already handles the no-command case
- Non-interactive mode without a command bails with a clear error message

## Test plan

- [x] `msb exec -t <sandbox>` attaches to default shell
- [x] `msb exec -t <sandbox> -- python` still works as before
- [x] `msb exec <sandbox> -- ls` still works as before
- [ ] Piping without a command (`echo | msb exec <sandbox>`) gives a clear error

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -260,6 +260,9 @@ jobs:
     name: Node.js SDK Tests
     needs: check
     runs-on: self-hosted-ubuntu-2404-x64
+    container:
+      image: ubuntu:24.04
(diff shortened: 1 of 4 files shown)


appcypher (approved):
LGTM

PR #545, opened 2026-04-13
fix(net): access localhost

wimpheling:
Bug

It is not possible to connect to services running on the host from within a microsandbox.

I was only able to fix it using an LLM, so I'm really not sure if this is a bug or feature, or if you would prefer a more generic way of fixing that issue.
Root cause

Two reasons:

    127.0.0.1 is the guest's own loopback. The guest kernel owns 127.0.0.0/8 on its lo interface, so TCP connections to 127.0.0.1 never leave via eth0 and never reach the smoltcp proxy.

    Connecting to the gateway IP fails silently. smoltcp accepts the SYN via any_ip=true, but the host-side proxy calls TcpStream::connect(gateway_ip:port) — a virtual IP that has no host interface — so the connection times out.

Fix

When the proxy is about to dial a connection destined for the gateway IP, rewrite it to 127.0.0.1/::1 before connecting. The gateway IP becomes a stable alias for the host, reachable from the guest via the default route.

From inside the sandbox, connect to the gateway IP (first nameserver in /etc/resolv.conf):

GW=$(grep nameserver /etc/resolv.conf | head -1 | awk '{print $2}')
curl http://$GW:<port>```


wimpheling:
## Bug

It is not possible to connect to services running on the host from within a microsandbox. 

I was only able to fix it using an LLM, so I'm really not sure if this is a bug or feature, or if you would prefer a more generic way of fixing that issue.

## Root cause

Two reasons:

1. **`127.0.0.1` is the guest's own loopback.** The guest kernel owns `127.0.0.0/8` on its `lo` interface, so TCP connections to `127.0.0.1` never leave via `eth0` and never reach the smoltcp proxy.

2. **Connecting to the gateway IP fails silently.** smoltcp accepts the SYN via `any_ip=true`, but the host-side proxy calls `TcpStream::connect(gateway_ip:port)` — a virtual IP that has no host interface — so the connection times out.

## Fix

When the proxy is about to dial a connection destined for the gateway IP, rewrite it to `127.0.0.1`/`::1` before connecting. The gateway IP becomes a stable alias for the host, reachable from the guest via the default route.

From inside the sandbox, connect to the gateway IP (first `nameserver` in `/etc/resolv.conf`):


```sh
GW=$(grep nameserver /etc/resolv.conf | head -1 | awk '{print $2}')
curl http://$GW:<port>```

**Comment by toksdotdev** [comment]
appreciate you giving it a shot :)

i don’t think we should treat the gateway IP as equivalent to host 127.0.0.1. 

127.0.0.1 is specifically for services that should only be reachable on the guest loopback and not exposed more broadly on the host network. Rewriting traffic sent to the sandbox gateway so it ultimately connects to host 127.0.0.1 deviates from that.

if an application sandbox needs to access a host service, they should do it through the gateway IP. you'd need explicitly allow traffic to a private subnet in microsandbox. below is an example (see [docs](https://docs.microsandbox.dev/sdk/rust/networking#networkpolicy-presets) for reference):

```rust
let policy = NetworkPolicy {
    default_action: Action::Deny,
    rules: vec![Rule::allow_outbound(Destination::Group(DestinationGroup::Private))],
};

let _sandbox = Sandbox::builder("private-egress")
   .image("alpine")
   .network(|n| n.policy(policy))
   .create()
   .await?;
```

if one still desires that loopback maps directly to host, a quick workaround could be to run a forwarder like `socat`, or modify the routing table within the sandbox to route loopback to the host gateway IP.

toksdotdev:
really appreciate the PR. opted to close this in favour of supporting `host.microsandbox.internal` hostname. this should go out with the upcoming v0.4 release. 

there are valid cases where clients within sandboxes do need to speak to applications bound to 127.0.0.1 (within the sandbox) instead of the host, so doing a 1:1 mapping of 127.0.01 to the gateway IP would violate such expectations.

PR #549, opened 2026-04-13
Add sandbox-wide guest rlimits

### LINKED ISSUE #497: Support VM-wide default_rlimits, not just per-exec
### Summary

Currently, rlimits are only applied per-exec command via `setrlimit()` in the child process after `fork()` (`crates/agentd/lib/session.rs`). Users want the ability to set default rlimits at the VM/sandbox level so that every process inherits them automatically.

### Current behavior

Rlimits must be specified on each exec invocation via the `exec_with` closure:

```rust
sb.exec_with("my-cmd", |e| {
    e.args(["arg"])
     .rlimit(RlimitResource::Nofile, 1024)
     .rlimit_range(RlimitResource::Nproc, 512, 4096)
}).await?;
```

### Desired behavior

Rlimits can be set at the sandbox level and apply to all processes:

```rust
SandboxBuilder::new("my-sandbox")
    .image("alpine:latest")
    .rlimit(RlimitResource::Nofile, 1024)
    .rlimit_range(RlimitResource::Nproc, 512, 4096)
    .build()
    .await?;
```

Per-exec rlimits should still work and override the defaults.

### Context

Many services expect system-wide resource limits to be set at init level rather than per-process.

enricoschaaf:
Closes #497.

## Summary

This adds sandbox-wide guest rlimits that are applied by `agentd` PID 1 during guest startup.

Today microsandbox only supports per-exec rlimits. That is too late for bootstrap scripts and long-lived daemons started during guest init, because they inherit PID 1's baseline limits before any per-exec override runs.

## What changed

- adds sandbox-wide `SandboxConfig::rlimits`
- adds `SandboxBuilder::rlimit(...)` and `SandboxBuilder::rlimit_range(...)`
- passes sandbox-wide rlimits into the guest via `MSB_RLIMITS`
- applies those limits in `agentd` before the rest of init runs
- keeps per-exec rlimits available so one-off execs can still override the baseline

## Validation

- `cargo test --manifest-path /tmp/msb-baseline/crates/agentd/Cargo.toml --lib`
- `cargo test --manifest-path /tmp/msb-baseline/crates/microsandbox/Cargo.toml --lib`

Sorry for the delay, let me know if anything is not like you want to happy to learn and adapt, hear to learn how the repo works


enricoschaaf:
ah sorry on nightly myself let me adapt to stable

enricoschaaf:
seems the node issue is rather an issue of CI than of the code?

toksdotdev:
@enricoschaaf fixing that in a moment :)

crates/microsandbox/lib/runtime/spawn.rs
```diff
@@ -380,6 +380,21 @@ fn push_file_mounts_spec(
     }
 }
 
+fn encode_rlimits(rlimits: &[Rlimit]) -> String {
```
toksdotdev, line 382: nit: doc comment.

crates/agentd/lib/init.rs
toksdotdev, line 1: there's already a `parse_rlimit_resource(..)` and `parse_rlimits(..)` implementation in `session.rs`. let's re-use that function.

crates/agentd/lib/init.rs
```diff
@@ -65,6 +66,58 @@ pub fn init() -> AgentdResult<()> {
...
+        let (soft, hard) = parse_rlimit_pair(limit_spec).map_err(|err| {
+            AgentdError::Init(format!(
+                "{ENV_RLIMITS} has invalid limit for {resource_name}: {err}"
+            ))
+        })?;
+
+        let limit = libc::rlimit {
+            rlim_cur: soft as libc::rlim_t,
+            rlim_max: hard as libc::rlim_t,
+        };
```
toksdotdev: per my other comment. you can reuse `parse_rlimits(..)`.

crates/agentd/bin/main.rs
```diff
@@ -18,6 +18,11 @@ fn main() {
     // Capture CLOCK_BOOTTIME immediately — this represents kernel boot duration.
     let boot_time_ns = microsandbox_agentd::clock::boottime_ns();
 
+    if let Err(e) = microsandbox_agentd::init::apply_rlimits() {
```
toksdotdev: nit: update the phases doc comments below (with this being first phase)

enricoschaaf:
Thanks! Will update

crates/agentd/lib/init.rs
enricoschaaf: see other comment

crates/agentd/lib/init.rs
enricoschaaf: It seemed like reusing session.rs was not giving me everything, so I looked how exec does it and tired to get the best reuse this way. Not sure if you like it let me know.

toksdotdev (approved):

toksdotdev (commented):
build is failing. will take a look later today.

appcypher (approved):
LGTM

PR #552, opened 2026-04-14
define and enforce max threshold for agentd boot sequence in CI 

eric-14:
# What does the PR solve. 

The PR updates github workflows(check.yml) to include verification of boot timings defined in the scripts/ci/boot-timing-thresholds.json file. The python script in scripts/ci directory consumes the JSON file and outputs ms and ns of the boot timings. 

# Further  additions. 

To test boot timings of different sandboxes an example sandbox has been created in the example directory. Modification can be done to the file main.rs in the boot-timing-ci directory to test boot times of different images.


appcypher:
im closing this because we couldn't get this in in time and we were conflicted on whether this is the right approach to the problem. apologies for taking so long to decide.

PR #562, opened 2026-04-16
build: use vendored dbus

Glatzel:
Explicitly use dbus crate and add the `vendored` feature, so that we do not need to install dbus using system package manager.


appcypher (approved):
LGTM

PR #569, opened 2026-04-17
execStreamWithConfig for TS SDK

wellsbunk5:
adding execStreamWithConfig for ts sdk so that it can handle piped stdin.

This is already available on Rust and Python SDK and the infra for it was already built out. Just need to make it available.


toksdotdev (approved):

toksdotdev:
made a few additions. but asides that, lgtm!

PR #571, opened 2026-04-17
DNS response mapping

wellsbunk5:
Problem:
Right now using node fetch from an alpine image fails for any website that does not have an AAAA record.
node fetch -> getaddrinfo -> musl -> parallel queries for A and AAAA records -> No AAAA records returns REFUSED EAI_FAIL for the entire lookup even though A record query is a success.

Fix:
Map the correct response from the query instead of just using REFUSED. Then it can still succeed since 1 query passes.
| Source                                             | RCODE    |
|----------------------------------------------------|----------|
| Successful lookup with answers                     | NOERROR  |
| `NoRecordsFound { response_code: NoError }`        | NOERROR (NoData) |
| `NoRecordsFound { response_code: NXDomain }`       | NXDOMAIN |
| `RequestRefused` from upstream                     | REFUSED  |
| Timeout / transport / unknown lookup error         | SERVFAIL |
| Local `blocked_domains` / suffix / rebind block    | REFUSED  |

### DIFF
--- a/crates/network/lib/dns/interceptor.rs
+++ b/crates/network/lib/dns/interceptor.rs
@@ -13,6 +13,9 @@ use std::collections::HashSet;
 use std::sync::Arc;
 
 use bytes::Bytes;
+use hickory_proto::op::ResponseCode;
+use hickory_proto::rr::Record;
+use hickory_proto::rr::rdata::SOA;
 use smoltcp::iface::SocketSet;
 use smoltcp::socket::udp;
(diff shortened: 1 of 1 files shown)


toksdotdev:
closing this in favour of https://github.com/superradcompany/microsandbox/pull/578. realised that `hickory-resolver` had some special behaviours than intended, which avoids such classes of issue completely. so made the switch over to `hickory-client`. 

thanks a lot for raising the issue and also a potential fix.

PR #577, opened 2026-04-19
ci: add boot and init time enforcement to CI

### LINKED ISSUE #473: define and enforce max threshold for agentd boot sequence in CI
we need to be very mindful the agentd boot sequence, and continue to keep it as minimal as possible. it's very tempting to keep adding things there. one of the selling points of microsandbox is how quick we are to spin up. we shouldn't trade that for just any features (unless we badly need to).

the first step to that is to already start recording the  `vm enter -> boot ⁠ time`, and `⁠ boot -> init ⁠ time` as part of the CI process, and enforcing a specific PR can't merged there's a significant jump, or it exceeds that max threshold.

Unique-Usman:
Add automated boot and init time performance testing to prevent regressions.
- Enforces 100ms thresholds for both boot and init times
- Fails CI build if thresholds are exceeded
- Uses existing timing infrastructure from agentd
- Cleans up test sandbox automatically
- Add init_time to agent client connection logs for visibility

---

Fixes #473

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -293,7 +293,7 @@ jobs:
           npm run build
 
   # ---------------------------------------------------------------------------
-  # Node.js SDK smoke tests (requires KVM)
+  # Node.js SDK smoke tests and Enforce Boot and Init Time(requires KVM)
   # ---------------------------------------------------------------------------
   node-sdk-test:
     name: Node.js SDK Tests
@@ -344,3 +344,8 @@ jobs:
           export PATH="$HOME/.microsandbox/bin:$PATH"
           export LD_LIBRARY_PATH="${{ github.workspace }}/build:$HOME/.microsandbox/lib"
           npm test
+      - name: Enforce Boot and Init Time
+        run: |
+          chmod +x scripts/check-boot-and-init-performance.sh
+          LD_LIBRARY_PATH="${{ github.workspace }}/build:$HOME/.microsandbox/lib" \
+            ./scripts/check-boot-and-performance.sh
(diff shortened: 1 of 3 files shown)


Unique-Usman:
Fix: https://github.com/superradcompany/microsandbox/issues/473

Unique-Usman:
Also, when I tested the script locally, it uses about 

Boot time: 36701211ns  ~ 37 milliseconds
Init time: 35198971ns ~ 35 milliseconds
Boot performance test passed!

I set the BOOT_TIME and INIT max threshold to 100ms, may it should be higher or lower than this @appcypher ? Also, instead of parsing from log, do you think supporting some config file that return json for the timings is better approach ?

toksdotdev:
this is heading in the right direction, but I don’t think log-scraping is the best long-term approach here. 

the timing data already exists in the structured `core.ready` payload, so it’d be cleaner and less brittle to expose/read that directly in CI rather than parsing RUST_LOG output. 

also, the workflow currently `calls check-boot-and-performance.sh`, but the script added is `check-boot-and-init-performance.sh`.

there's also a strong overlap between this PR and https://github.com/superradcompany/microsandbox/pull/552 (which i also have lots of reservations about).

Unique-Usman:
> erlap between



> this is heading in the right direction, but I don’t think log-scraping is the best long-term approach here.
> 
> the timing data already exists in the structured `core.ready` payload, so it’d be cleaner and less brittle to expose/read that directly in CI rather than parsing RUST_LOG output.
> 
Yeah, that is true.
> also, the workflow currently `calls check-boot-and-performance.sh`, but the script added is `check-boot-and-init-performance.sh`.
mistake, I am sorry for that.
> 
> there's also a strong overlap between this PR and #552 (which i also have lots of reservations about).

Oops, I think this is a duplicate then, I am happy to close this unless @eric-14 does not want to continuing working on it, which I do not think so.

toksdotdev:
yeah, i'd rather have the tests in rust, and #552 is much closer aligned to that than yours. so i'll proceed to close this. appreciate the PR though. 🙌🏽

PR #583, opened 2026-04-20
feat(network): trust host CAs inside sandbox guests (opt-in)

toksdotdev:
## summary

- outbound https inside a sandbox fails whenever the host is behind a tls-inspecting proxy (warp zero trust, zscaler, netskope, etc.) because the proxy's ca is trusted on the host but not in the guest.
- added an opt-in path that copies the host's trusted roots into the guest at boot, so the guest trusts whatever the host trusts.
- off by default. enable via `.trust_host_cas(true)` in rust, `trustHostCas: true` in the node sdk, `trust_host_cas=True` in the python sdk, or `--trust-host-cas` on the cli.

## test plan

- [x] `cargo test -p microsandbox-network tls::host_cas`
- [x] `cargo check --workspace` clean
- [x] macos + warp on + `--trust-host-cas`: `msb run alpine -- apk update` succeeds
- [x] macos + warp on, no flag: original failure reproduces
- [x] linux: apk/curl unchanged

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2674,6 +2674,7 @@ dependencies = [
  "lru",
  "microsandbox-utils",
  "msb_krun",
+ "pem",
  "rcgen",
  "resolv-conf",
  "rustls",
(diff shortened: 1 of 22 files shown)


toksdotdev:
block on this. this should be disabled by default. if users need it, they should opt-in manually.

PR #587, opened 2026-04-20
feat(sdk/go): Go SDK with full API parity (#568)

### LINKED ISSUE #568: feat: Go SDK
# Summary

Create a Go SDK that provides the same capabilities as the existing Node.js/TypeScript SDK,
adapted to Go conventions. Go is the dominant language for infrastructure and DevOps tooling —
the exact microsandbox audience — so this is a high-priority SDK.

## API design

The design is meant to be idiomatic:

```go
import "github.com/superradcompany/microsandbox/sdk/go"

sb, err := microsandbox.NewSandbox(ctx, "my-sandbox",
    microsandbox.WithImage("python"),
    microsandbox.WithMemory(512),
    microsandbox.WithCPUs(2),
    microsandbox.WithEnv(map[string]string{"PYTHONDONTWRITEBYTECODE": "1"}),
    microsandbox.WithWorkdir("/app"),
    microsandbox.WithVolume("/app/src", microsandbox.BindVolume("./src", microsandbox.Readonly())),
    microsandbox.WithNetwork(microsandbox.PublicOnly()),
)

output, err := sb.Exec(ctx, "python", []string{"-c", "print('Hello!')"})
fmt.Println(output.Stdout())

err = sb.Stop(ctx)
```

## Core API surface

**Sandbox** — lifecycle, execution, filesystem, metrics

- `NewSandbox(ctx, name, opts...)` — create and boot
- `NewSandboxDetached(ctx, name, opts...)` — survives parent process
- `GetSandbox(ctx, name)` / `ListSandboxes(ctx)` / `RemoveSandbox(ctx, name)`
- `sb.Exec(ctx, cmd, args, opts...)` — collected output
- `sb.ExecStream(ctx, cmd, args, opts...)` — streaming via channel
- `sb.Shell(ctx, script, opts...)`
- `sb.FS()` — `Read`, `Write`, `List`, `Stat`, `CopyIn`, `CopyOut`
- `sb.Metrics(ctx)` — CPU, memory, disk I/O, network I/O
- `sb.Stop(ctx)` / `sb.Kill(ctx)` / `sb.Detach(ctx)`

**Volume** — persistent storage

- `NewVolume(ctx, name, opts...)` / `ListVolumes(ctx)` / `RemoveVolume(ctx, name)`
- Mount factories: `BindVolume(path)`, `NamedVolume(name)`, `TmpfsVolume(sizeMiB)`

**NetworkPolicy** — presets

- `NoNetwork()` / `PublicOnly()` / `AllowAll()` / `Allowlist(hosts...)`

**Secret** — host-side credential injection

- `EnvSecret(name, opts...)`

**Patch** — pre-boot rootfs modifications

- `TextPatch(path, content)` / `MkdirPatch(path, mode)` / `AppendPatch(path, content)`

**Streaming** — channel-based event loop

```go
handle, err := sb.ExecStream(ctx, "tail", []string{"-f", "/var/log/app.log"})
for event := range handle.Events() {
    switch e := event.(type) {
    case microsandbox.StdoutEvent:
        os.Stdout.Write(e.Data)
    case microsandbox.ExitedEvent:
        break
    }
}
```

**Utilities**

- `AllSandboxMetrics(ctx)` — metrics for all running sandboxes
- Typed errors with `IsKind` / `IsSandboxNotFound` / `IsExecTimeout` / `IsExecFailed`

## Implementation approach

Use CGO to wrap the microsandbox Rust crate as a native Go extension, via a new `sdk/go-ffi/`
Rust crate that exposes `extern "C"` blocking wrappers (with an internal Tokio runtime).
Same strategy as the Node.js SDK uses with NAPI and the Python SDK with PyO3/maturin.

I have a working implementation of the full API surface ready — happy to open a PR once
this direction looks good.

khrees:
Closes #568

## Summary

Full Go SDK for microsandbox, implemented via CGO over a new `sdk/go-ffi` Rust cdylib. The library is loaded at runtime via `dlopen`/`dlsym` — no build-time linking required. API design follows Go conventions (functional options, context-first, typed errors) while covering the same surface as the Node and Python SDKs.

## What's included

**Sandbox lifecycle**
- `CreateSandbox` / `CreateSandboxDetached`, `GetSandbox`, `StartSandbox`, `StartSandboxDetached`, `ListSandboxes`, `RemoveSandbox`
- `SandboxHandle` — lightweight DB record reference with `Connect`, `Start`, `Stop`, `Kill`, `Remove`, `Metrics`
- Live `Sandbox` — `Stop`, `StopAndWait`, `Kill`, `Wait`, `Drain`, `Detach`, `Close`, `OwnsLifecycle`, `RemovePersisted`
- `Attach` / `AttachShell` for interactive PTY sessions
- `AllSandboxMetrics`, `Sandbox.MetricsStream(interval)`

**Exec**
- `Exec` / `Shell` — collected output
- `ExecStream` / `ShellStream` — streaming with `ExecHandle.Recv`, `Collect`, `Wait`, `Kill`, `Signal`, `ID`, `TakeStdin`
- Per-exec `User` and `Env` overrides via `WithExecUser` / `WithExecEnv`

**Filesystem**
- `SandboxFs` — `Read`, `Write`, `List`, `Stat`, `CopyFromHost`, `CopyToHost`, `Mkdir`, `Remove`, `RemoveDir`, `Copy`, `Rename`, `Exists`
- Streaming: `ReadStream` (`io.WriterTo`) and `WriteStream` (`io.Writer`)

**Volumes**
- `CreateVolume`, `GetVolume`, `ListVolumes`, `RemoveVolume`
- `Volume.Path()` / `Volume.FS()` and `VolumeHandle` with full metadata + `FS()`
- `VolumeFs` — direct host-side file ops on the volume directory

**Options & factories**
- `NetworkPolicy.None/PublicOnly/AllowAll`, custom rules
- `Secret.Env(...)` for proxy-layer credential injection
- `Patch.Text/Append/Mkdir/Remove/Symlink/CopyFile/CopyDir`
- `Mount.Bind/Named/Tmpfs` with `WithMounts`
- `WithHostname`, `WithUser`, `WithReplace` on sandbox creation

**FFI layer** (`sdk/go-ffi/`)
- New Rust cdylib exposing `extern "C"` async wrappers with an internal Tokio runtime
- Handle registry pattern (opaque `u64` keys) for sandbox, exec, metrics stream, fs read/write handles
- JSON over FFI boundary; cancel via shared `AtomicBool` keyed by cancel ID

## Test plan
- [ ] `go build ./...` — passes
- [ ] `go vet ./...` — passes
- [ ] Integration tests in `sdk/go/*_test.go` against a running microsandbox daemon
- [ ] Verify `EnsureInstalled` downloads and loads the cdylib correctly on a fresh machine
- [ ] Test detached sandbox survives Go process exit and reattaches via `GetSandbox`


appcypher:
Bro. Thanks for contributing this. I already took a look at it a couple days back. 
I will do a more thorough review over the weekend.

toksdotdev:
@Khrees2412 is the Go implementation auto-generated? or just the `ffi.go`?

asking in order to guage how much work it'd be to maintain.

khrees:
@toksdotdev 

Mostly handwritten. ffi.go is the only really low-level FFI bridge layer; the rest of the Go SDK is normal handwritten Go.

Claude helped a lot with the initial implementation, but it’s not codegen-driven. Maintenance should mostly just be keeping `ffi.go` in sync with Rust FFI changes, plus the usual wrapper/test updates when the API changes.

zTrix:
May I ask is there any expectation time that this PR will be merged? I'm currently working on a go program and I'd like to use microsandbox embeded.

toksdotdev:
@zTrix this is top of my list atm, hopefully this weekend or early next week.

@khrees ok if i push changes to your branch as i walk my way through the changes (if any)? i feel it'd be faster than multiple back and forths for reviews. let me know what you think.

cedws:
Hey, I had a look as I've been keenly waiting for a Go SDK too. The API is clean and idiomatic, and supporting back to Go 1.21 is nice decision. When it comes to FFI-based wrappers I'm usually a bit cautious as they're bug prone and can be slow, but in this case I think it's fine, and at least it provides some consistency between SDKs.

Only one thing I've found so far I wanted to ask about: in the case that the consumer tries to create a duplicate sandbox with the `WithReplace()` option before shutting down the first, is it expected that the second `CreateSandbox` will block until the context is canceled?

For example, this code blocks forever on the second iteration:

```go
func main() {
	if err := msb.EnsureInstalled(context.Background()); err != nil {
		panic(err)
	}

	for i := range 2 {
		sandbox, err := msb.CreateSandbox(
			context.Background(),
			"test",
			msb.WithCPUs(1),
			msb.WithReplace(),
			msb.WithImage("alpine"),
		)
		if err != nil {
			panic(err)
		}
		log.Println("created")
		outp, err := sandbox.Exec(context.Background(), "echo", []string{"xyz"})
		if err != nil {
			panic(err)
		}

		log.Printf("VM %d: %s", i, outp.Stdout())
	}
}
```

Should `CreateSandbox` return something like `ErrAlreadyCreated` if a sandbox with the same name is already running?

khrees:
@toksdotdev Yes, that's fine


@cedws Fair point but it doesn't block forever; there's a 30s internal timeout in the core Rust lib `(crates/microsandbox/lib/sandbox/mod.rs:1536)`. The caller should pass a context with a timeout, which is standard Go practice. That said, we could also wrap an internal timeout in the Go SDK as a safety net but what then would be a safe duration?

cedws:
Ah ok, I didn't know about the timeout in the core library, I've only read this PR. 

```
sandbox still running: cannot replace sandbox 'test': existing sandbox did not stop in time
```

I think this is sufficient. Do you think there'll ever be a case a consumer would want to wait *longer* than 30s?

khrees:
I don't think so, 30s is plenty. If a process doesn't stop after SIGTERM + 30s, it's stuck. Waiting longer won't help; you'd need to kill it.

toksdotdev:
getting back to this. i could see 30s being excessive in quite a lot of cases. made a few changes:
- `replace()`: uses default timeout of 10s
- `replace_with_grace(duration)`: allows configuring a timeout.

i'll also proceed to merge this in now. follow up work will be to bring the Go docs up-to-date.

toksdotdev (approved):
really appreciate the effort on this @khrees 🫡

PR #589, opened 2026-04-20
feat(sdk/node-ts): add createWithProgress for image pull events

### LINKED ISSUE #565: logLevel in Sandbox.create doesn't log anything
Using the typescript sdk, doing `Sandbox.create` show no logs if logLevel is set. Even when using alpine, it "creates" the sandbox, yet no logs are shown.

My intention was to view the image pull progress if any; but logLevel doesn't not give ANY information.

Eronmmer:
## Summary

Adds `Sandbox.createWithProgress` and `Sandbox.createDetachedWithProgress` to the Node TS SDK. They return a `PullSession` that streams image pull events and resolves to the live sandbox.

```js
const session = await Sandbox.createWithProgress({
  name: "my-sb",
  image: "python:3.12-alpine",
  pullPolicy: "always",
});

for await (const ev of session) {
  if (ev.eventType === "layer_download_progress") {
    console.log(`L${ev.layerIndex}: ${ev.downloadedBytes}/${ev.totalBytes ?? "?"}`);
  }
}

const sandbox = await session.result();
```

`PullSession` exposes `recv()`, result(), and `Symbol.asyncIterator`. The `PullEvent` shape covers all 14 variants of the Rust `PullProgress` enum (resolving, resolved, layer_download_, layer_materialize_, stitch_*, complete).

Also clarifies the `logLevel` JSDoc, since that field was the usual source of confusion: it controls the msb supervisor's own stderr, not anything the SDK surfaces.

Docs: added reference entries for `createWithProgress`, `createDetachedWithProgress`, `PullSession`, and `PullEvent` in `docs/sdk/typescript/sandbox.mdx`.

Closes #565


sdk/node-ts/lib/sandbox.rs
```diff
@@ -440,6 +494,298 @@ impl AsyncGenerator for JsMetricsStream {
...
+    ) -> Self {
+        // Bridge the PullProgress channel into a local PullEvent channel,
+        // matching the pattern used by metrics_stream. The extra hop lets
+        // AsyncGenerator::next yield napi-compatible events without a match
+        // on every recv.
+        let (tx, rx) = tokio::sync::mpsc::channel::<PullEvent>(64);
+        tokio::spawn(async move {
+            while let Some(event) = handle.recv().await {
+                if tx.send(convert_pull_progress(event)).await.is_err() {
+                    break;
```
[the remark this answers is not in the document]
Eronmmer: Good catch, applied. switched to `try_send` with a 256-slot channel to match the upstream `PullProgressSender` semantics

Eronmmer:
Closes #565

docs/sdk/typescript/sandbox.mdx
```diff
@@ -339,6 +383,41 @@ Async stream for receiving periodic metrics snapshots.
...
+| `'layer_materialize_complete'` | `layerIndex`, `diffId` |
+| `'stitch_merging_trees'` | `layerCount` |
+| `'stitch_writing_fsmeta'` | (none) |
+| `'stitch_writing_vmdk'` | (none) |
+| `'stitch_complete'` | (none) |
+| `'complete'` | `reference`, `layerCount` |
+
+| Field | Type | Description |
+|-------|------|-------------|
+| eventType | `string` | Variant discriminator |
```
toksdotdev: can this be a ref, `PullEventType`. and the table prior under its own section, `PullEventType`.

docs/sdk/typescript/sandbox.mdx
```diff
@@ -54,6 +54,50 @@ Create and boot in detached mode. The sandbox survives after your process exits.
 
 ---
 
+#### Sandbox.createDetachedWithProgress()
```
toksdotdev: we don't expose both of these functions for rust. could we do that as well. ok with doing it another PR.

docs/sdk/typescript/sandbox.mdx
```diff
@@ -54,6 +54,50 @@ Create and boot in detached mode. The sandbox survives after your process exits.
...
+| Type | Description |
+|------|-------------|
+| [`PullSession`](#pullsession) | Async iterable of pull events. Call `result()` to obtain the running sandbox. |
+
+---
+
+#### Sandbox.createWithProgress()
+
+```typescript
+static createWithProgress(config: SandboxConfig): Promise<PullSession>
```
toksdotdev: let's rename `PullSession` to `PullProgress` for consistency with that of python.

docs/sdk/typescript/sandbox.mdx
```diff
@@ -339,6 +383,41 @@ Async stream for receiving periodic metrics snapshots.
 | [Symbol.asyncIterator] | `AsyncGenerator<`[`SandboxMetrics`](#sandboxmetrics)`>` | Use with `for await...of` |
 | recv() | `Promise<`[`SandboxMetrics`](#sandboxmetrics)` \| null>` | Receive next snapshot. Returns `null` when the stream ends. |
 
+### PullEvent
```
toksdotdev: can this also be documented for python. it's missing as well.

sdk/node-ts/lib/sandbox.rs
```diff
@@ -440,6 +494,299 @@ impl AsyncGenerator for JsMetricsStream {
...
+            manifest_digest: Some(manifest_digest.to_string()),
+            layer_count: Some(layer_count as u32),
+            total_download_bytes: total_download_bytes.map(|b| b as f64),
+            layer_index: None,
+            digest: None,
+            diff_id: None,
+            downloaded_bytes: None,
+            total_bytes: None,
+            bytes_read: None,
+        },
```
toksdotdev: any reason we can't return `PullProgress` directly? i'm thinking, annotating it with something like `#[serde(tag = "event", content = "data")]`. is this possible?

this would avoid the need for a new struct type.

docs/sdk/typescript/sandbox.mdx
Eronmmer: done

docs/sdk/typescript/sandbox.mdx
Eronmmer: will create a follow up issue for this

docs/sdk/typescript/sandbox.mdx
Eronmmer: checked Python first to make sure I'd land the rename in the right direction. Python already has both names: `PullSession` is the class (`_microsandbox.pyi:216`) and `PullProgress` is the event union type alias (`events.py:101`). Our TS `PullSession` already matches Python's class name. renaming it to `PullProgress` would introduce a new mismatch and collide with Python's event type.

I think what would actually align the two SDKs is renaming our TS `PullEvent` to  `PullProgress` (which is Python's name for the event type). That also slots into the Python doc change from your comment below this one, where I already used PullProgress as the section name. Does that match your intent?

docs/sdk/typescript/sandbox.mdx
Eronmmer: check the comment above

sdk/node-ts/lib/sandbox.rs
Eronmmer: ran some scripts to verify the way it works — napi-rs 3 handles tagged enums cleanly. Dropped a throwaway `#[napi]` enum into `sdk/node-ts/lib/` to see what gets emitted:

```rust
#[napi]
pub enum ExperimentalPullProgress {
    Resolving { reference: String },
    Resolved { reference: String, manifest_digest: String, layer_count: u32, total_download_bytes: Option<f64> },
    LayerDownloadProgress { layer_index: u32, digest: String, downloaded_bytes: f64, total_bytes: Option<f64> },
    Complete { reference: String, layer_count: u32 },
}
```

produces in `index.d.cts`:

```ts
export type ExperimentalPullProgress =
  | { type: 'Resolving', reference: string }
  | { type: 'Resolved', reference: string, manifestDigest: string, layerCount: number, totalDownloadBytes?: number }
  | { type: 'LayerDownloadProgress', layerIndex: number, digest: string, downloadedBytes: number, totalBytes?: number }
  | { type: 'Complete', reference: string, layerCount: number }
```

A proper discriminated union — TS narrows per variant, required/optional fields are correct per variant. Strictly nicer than the current `PullEvent` flat struct where every field is optional and you have to switch on a string tag manually.

Two notes:

1. `#[serde(tag=..., content=...)]` turns out to be the wrong knob here — napi-rs does its own Rust→JS translation independent of serde. But we don't need serde; `#[napi]` on the enum alone produces what we want.
2. I don't think we want `#[napi]` directly on `microsandbox_image::PullProgress` — its fields are `Arc<str>` (napi-rs doesn't handle that), and it feels inverted to pull napi-rs into the core crate for one SDK. So this would be a local `#[napi]` enum in `sdk/node-ts/lib` mirroring the core variants, with the converter doing `PullProgress::Resolving { reference } → LocalEnum::Resolving { reference: reference.to_string() }`. The converter actually shrinks — one variant → one variant, no 11-field struct literal.

lmk how you'd like to proceed, PS, i'm not deep in Rust yet so i'm open to other ideas or if you have something else in mind

toksdotdev: thanks for looking into this. yes, i like the discriminant sample that you shared. let'd do that only nit, could we make the type snake instead of pascal cased?

1. yeah, that make sense. the serde was just more of an example.
2. sounds like a good plan. a local pull progress enum sounds right to me.

docs/sdk/typescript/sandbox.mdx
toksdotdev: make sense. `PullEvent` -> `PullProgress` sounds about right to me. yes, let's do that.

docs/sdk/typescript/sandbox.mdx
toksdotdev: seen. thanks 🙏🏽

docs/sdk/typescript/sandbox.mdx
Eronmmer: done.

sdk/node-ts/lib/sandbox.rs
Eronmmer: Pushed the refactor. Local `PullProgress` enum replaces the flat `PullEvent` struct (with `discriminant_case = "snake_case"` for the tags you asked for), converter collapses to one-line-per-variant, `PullEvent → PullProgress` rename lands in the same change. Also synced `index.d.ts` so both generated type files agree. and all 18 smoke tests pass

docs/sdk/python/sandbox.mdx
```diff
@@ -499,9 +499,41 @@ Controls when the SDK fetches an OCI image from the registry.
...
+| `LayerExtractProgress` | `layer_index: int`, `bytes_read: int`, `total_bytes: int` |
+| `LayerExtractComplete` | `layer_index: int`, `diff_id: str` |
+| `LayerIndexStarted` | `layer_index: int` |
+| `LayerIndexComplete` | `layer_index: int` |
+| `PullComplete` | `reference: str`, `layer_count: int` |
+
+Narrow with `isinstance`:
+
+```python
+from microsandbox import Sandbox, Resolved, LayerDownloadProgress
```
toksdotdev, line 784: huh. can you confirm that an enum isn't emitted for py? per this line, it reads to me as though each of the variants are dedicated type, which i'd hope not to be so 🤔

Eronmmer: Confirmed. no enum. Each variant is a frozen dataclass; `PullProgress` at events.py:101 is just `Resolving | Resolved | ...` (a union alias). that's how the Python SDK already did it

Eronmmer:
@toksdotdev just a heads up that this pr was open before the ts sdk rewrite. there's now a couple of merge conflicts

toksdotdev:
@Eronmmer could you help resolve the conflicts. thanks.

toksdotdev (approved):

PR #599, opened 2026-04-23
feat(sdk): expose per-sandbox libkrunfw_path override

appcypher:
## Summary
- Add `libkrunfw_path: Option<PathBuf>` to `SandboxConfig` and a fluent `.libkrunfw_path(path)` method on `SandboxBuilder`, with file-existence validation at build time.
- In `spawn_sandbox`, prefer the per-sandbox override; fall back to `config::resolve_libkrunfw_path()` when unset. Not persisted (`#[serde(skip)]`) — libkrunfw is a host-side resource, not sandbox state.
- Surface the same option through the Node SDK as `libkrunfwPath?: string` on `SandboxConfig` (plus both `index.d.ts` / `index.d.cts`) and through the Python SDK as a `libkrunfw_path=...` kwarg on `Sandbox.create(...)`.
- Motivation: let SDK users point a specific sandbox at an alternative firmware bundle (e.g. unreleased libkrunfw during development) without mutating global config or env vars.
- Housekeeping: remove the dormant `_runtime.libkrunfw_path()` helper and its `MICROSANDBOX_LIBKRUNFW_PATH` env var in the Python wheel — the helper was never called and the Rust resolver never read that env var. libkrunfw is located correctly today via `MSB_PATH` + the resolver's `../lib/` fallback (matches the wheel bundle layout).

## Test Plan
- [x] `cargo check -p microsandbox -p microsandbox-node -p microsandbox-py` compiles cleanly.
- [x] `cargo test -p microsandbox` passes — existing builder/config/spawn tests unaffected.
- [x] Rust SDK: `Sandbox::builder("x").image("alpine").libkrunfw_path("/nonexistent").build()` returns `InvalidConfig` ("libkrunfw_path does not exist: ...").
- [x] Rust SDK: `.libkrunfw_path("/valid/libkrunfw.dylib")` is the path actually passed via `--libkrunfw-path` to the `msb sandbox` subprocess (override wins over global resolver).
- [x] Rust SDK: omitting `.libkrunfw_path(...)` continues to resolve via the existing global logic (no behavior change).
- [x] Node SDK: passing `{ libkrunfwPath: "/path/to/libkrunfw.dylib" }` in `SandboxConfig` forwards to the Rust builder and a sandbox created with it boots.
- [x] Python SDK: `await Sandbox.create("x", image="alpine", libkrunfw_path="/path/to/libkrunfw.dylib")` forwards the kwarg and boots with the override.
- [x] Python wheel: after removing the dormant helper, `import microsandbox` still works and `Sandbox.create(...)` resolves libkrunfw from the bundled `_bundled/lib/` (sibling of `_bundled/bin/msb`).


toksdotdev:
referencing https://github.com/superradcompany/microsandbox/issues/600 here.

i was thinking, we could allow folks override this via the global config.json. in which case, every sandbox starts with their custom kernel.

toksdotdev (approved):

appcypher:
> i was thinking, we could allow folks override this via the global config.json. in which case, every sandbox starts with their custom kernel.

that's always been supported. global config [already takes a libkrunfw path](https://github.com/superradcompany/microsandbox/blob/cafd73313380a254e5012858c3592c61b4009af7/crates/microsandbox/lib/config/mod.rs#L98)

PR #602, opened 2026-04-23
feat(network): add host.microsandbox.internal alias

### LINKED ISSUE #546: Cannot access host services from inside the sandbox
## Bug

It is not possible to connect to services running on the host from within a microsandbox.

I used an LLM to generate a fix, so it works but I'm not sure if it is the best way to integrate it in the lib (should this be more generic ? activable through an option as it may be a security issue to activate it by default ... )

## Root cause

Two reasons:

1. **`127.0.0.1` is the guest's own loopback.** The guest kernel owns `127.0.0.0/8` on its `lo` interface, so TCP connections to `127.0.0.1` never leave via `eth0` and never reach the smoltcp proxy.

2. **Connecting to the gateway IP fails silently.** smoltcp accepts the SYN via `any_ip=true`, but the host-side proxy calls `TcpStream::connect(gateway_ip:port)` — a virtual IP that has no host interface — so the connection times out.

## Fix

When the proxy is about to dial a connection destined for the gateway IP, rewrite it to `127.0.0.1`/`::1` before connecting. The gateway IP becomes a stable alias for the host, reachable from the guest via the default route.

From inside the sandbox, connect to the gateway IP (first `nameserver` in `/etc/resolv.conf`):

```sh
GW=$(grep nameserver /etc/resolv.conf | head -1 | awk '{print $2}')
curl http://$GW:<port>
```
see https://github.com/superradcompany/microsandbox/pull/545

toksdotdev:
## problem

guests couldn't reach host-side services from inside the sandbox without hacky solutions.

## solution

added `host.microsandbox.internal` which automatically resolves to the per-sandbox gateway ip. agentd was modified to write the mapping to `/etc/hosts`. in addition, a `DestinationGroup::Host` was added, and devs can now explicitly allow traffic to the host machine.

closes https://github.com/superradcompany/microsandbox/pull/545
fixes https://github.com/superradcompany/microsandbox/issues/546

### DIFF
--- a/crates/network/lib/dns/forwarder.rs
+++ b/crates/network/lib/dns/forwarder.rs
@@ -33,7 +33,8 @@ use bytes::Bytes;
 use futures::StreamExt;
 use hickory_client::client::Client;
 use hickory_client::proto::op::{Message, MessageType, ResponseCode};
-use hickory_client::proto::rr::{RData, RecordType};
+use hickory_client::proto::rr::rdata::{A, AAAA};
+use hickory_client::proto::rr::{RData, Record, RecordType};
 use hickory_client::proto::serialize::binary::{BinDecodable, BinEncodable};
 use hickory_client::proto::xfer::{DnsHandle, DnsRequest};
 use tokio::sync::{OnceCell, watch};
@@ -45,6 +46,7 @@ use super::common::transport::Transport;
 use super::nameserver::{read_host_dns_servers, resolve_nameservers};
 use crate::policy::NetworkPolicy;
 use crate::shared::{ResolvedHostnameFamily, SharedState};
+use crate::stack::GatewayIps;
 
 //--------------------------------------------------------------------------------------------------
 // Constants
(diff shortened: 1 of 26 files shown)


appcypher:
This [issue](https://github.com/superradcompany/microsandbox/issues/546) probably can be closed by this.

Also this [PR](https://github.com/superradcompany/microsandbox/pull/545) has been superseded by the general networking fixes we made recently including this one.

crates/network/lib/dns/forwarder.rs
```diff
@@ -410,6 +458,64 @@ fn build_status_response(query: &Message, rcode: ResponseCode) -> Option<Bytes>
...
+    qtype: RecordType,
+) -> Option<Bytes> {
+    let question = query.queries().first()?;
+    let name = question.name().clone();
+
+    let rdata = match qtype {
+        RecordType::A => RData::A(A::from(gateway.ipv4)),
+        RecordType::AAAA => RData::AAAA(AAAA::from(gateway.ipv6)),
+        _ => return None,
+    };
```
appcypher, line 533: Any idea why we map A to ipv4 and AAAA to ipv6?

toksdotdev: we want to return an ipv4 gateway address if the DNS query type is for IPv4.

PR #606, opened 2026-04-23
fix(network): send RST on upstream connect failure instead of FIN

toksdotdev:
## problem

when a proxy task's `TcpStream::connect` to the host-side upstream fails (e.g. the guest dialed a port where nothing is listening), the smoltcp socket currently closes cleanly with FIN. the guest sees "connection accepted then closed" rather than "connection refused", and happy-eyeballs clients that picked the wrong family stall instead of falling back.

## solution

tracked a per-connection `upstream_connected: Arc<AtomicBool>` that the proxy task flips to `true` after a successful upstream connect. when the tracker detects a dead proxy task with the flag still `false`, it aborts the smoltcp socket (RST) instead of closing it (FIN). wired through the plain tcp proxy, the tls bypass relay, and the tls intercept relay. dns / dot tasks skip the flag since they never produce a guest-visible unreachable upstream (servfail is synthesised locally).

## test plan

- [x] `cargo check --workspace --tests` clean
- [ ] manual: start a sandbox with `allow_all`, run `nc -vz host.microsandbox.internal <port-with-no-listener>`; expect immediate connection-refused rather than "succeeded" followed by a silent close
- [ ] manual: same but with a live listener on the host; expect normal clean close (FIN, not RST)

## note

stacked on top of #602 (uses `host.microsandbox.internal` as the easiest manual repro); can be rebased onto main after that merges.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2678,6 +2678,7 @@ dependencies = [
  "ipnetwork",
  "libc",
  "lru",
+ "microsandbox-protocol",
  "microsandbox-utils",
  "msb_krun",
  "pem",
--- a/crates/agentd/lib/config.rs
+++ b/crates/agentd/lib/config.rs
@@ -17,8 +17,8 @@ use std::env;
 use std::net::{Ipv4Addr, Ipv6Addr};
 
 use microsandbox_protocol::{
-    ENV_BLOCK_ROOT, ENV_DIR_MOUNTS, ENV_FILE_MOUNTS, ENV_HOSTNAME, ENV_NET, ENV_NET_IPV4,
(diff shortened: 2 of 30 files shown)


toksdotdev:
folded into #621.

the unique commit (ea9a46df, RST-on-upstream-connect-failure in `proxy.rs` + `conn.rs`) was cherry-picked onto `spec/network-policy-ingress` and is now part of #621. the rest of this branch was older unrebased history that already landed on main under different shas (host.microsandbox.internal alias, networking docs split, dns interceptor, integration test infra).

#621 covers both rejection paths in one place: ingress-policy-deny (publisher.rs, RST to peer) and egress-upstream-connect-failure (proxy.rs/conn.rs, RST to guest).

PR #654, opened 2026-05-03
Metrics Sampling Interval

wellsbunk5:
Add the ability to configure the metrics sample interval in ms from the config or override in the sdk command to create a sandbox.

This is so that users can configure how frequently metric samples are collected. It could be faster or slower but the default remains 1 second or 1000 ms. If a user wants to disable sampling completely they set it to 0 and the entire metric sampling task is skipped.

If they disable metric sampling by setting the interval to 0 and then they try to read the metrics from cli/sdk, they will receive and error saying metrics are disabled for that sandbox.

Example config:
```json
{
  "database": {
    "max_connections": 1
  },
  "sandbox_defaults": {
    "metrics_sample_interval_ms": 0
  }
}
```


crates/microsandbox/lib/config/mod.rs
```diff
@@ -131,6 +134,9 @@ pub struct SandboxDefaults {
 
     /// Default working directory inside the sandbox.
     pub workdir: Option<String>,
+
+    /// Default metrics sampling interval in milliseconds. `0` disables sampling entirely.
+    pub metrics_sample_interval_ms: u64,
```
toksdotdev: nit: 

could we use an `Option<NonZero<u64>>` instead, where `None` means metrics is disabled. `NonZero::new(..)` returns an `Option` if the value is `0`. it should help avoid a lot of the `== 0` checks.

crates/cli/lib/sandbox_cmd.rs
```diff
@@ -72,6 +72,14 @@ pub struct SandboxArgs {
...
     pub memory_mib: u32,
 
+    /// Metrics sampling interval in milliseconds; pass `null` to disable.
+    #[arg(
+        long = "metrics-sample-interval-ms",
+        default_value = "1000",
+        value_parser = parse_metrics_sample_interval,
+    )]
+    pub metrics_sample_interval_ms: Option<std::num::NonZero<u64>>,
+
```
toksdotdev: we could tweak this a little bit, and use two fields:
- `--metrics-sample-interval-ms`: 0 maps to `None`. interval > 0 maps to  `Some(interval)`. if unspecified, use default of 1000.
- `--disable-metrics-sample`: maps to `None`. if specified, it overrides the `--metrics-interval-ms`.

that should avoid us needing to parse `null` / `none`.

docs/configuration.mdx
```diff
@@ -94,6 +94,24 @@ Defaults applied to every sandbox unless overridden per-sandbox.
...
+often callers poll the DB.
+
+<Note>
+  Set the field to `null` to disable sampling. Omitting the field uses
+  the default (`1000` ms). Any positive integer enables sampling at
+  that interval; `0` is not accepted in JSON — use `null` instead.
+  Very small intervals are valid but produce a heavy stream of
+  `sandbox_metric` rows; intervals below `500` ms are discouraged in
+  production.
+</Note>
```
toksdotdev: with the 2 fields, i think we can simplify this doc to just a line each for each flag. we can then go into more depth with metrics sampling on the sampling page when we re-author metrics collection.

crates/microsandbox/lib/sandbox/metrics.rs
```diff
@@ -59,24 +66,40 @@ impl Sandbox {
...
+                    if disabled {
+                        if emitted {
+                            return None;
+                        }
+                        emitted = true;
+                        return Some((
+                            Err(MicrosandboxError::MetricsDisabled(sandbox_name)),
+                            (ticker, emitted, disabled),
+                        ));
+                    }
```
toksdotdev: given the disabled value remains unchanged, how about we yield much earlier vs checking on every tick. we could instead return `stream::once(async move { Err(MetricsDisabled(name)) }).left_stream()`. 

we might need to call `.right_stream()` on the result of `stream::unfold(...)

toksdotdev (approved):

PR #665, opened 2026-05-04
Optimize Image persistance queries

wellsbunk5:
The biggest change in this PR is to add a fast path for the Image upserts that happen on sandbox creation. If a image is reused and already cached there should be no need to upsert all image layers.  Currently the upserts happen everytime a sandbox is created. This PR checks if the Image already exists and if the manifest digest matches. If the manifest digest matches there is no point to upsert all rows and instead does a single update last_used_at column statement. For cached images it removes 25~30 statements  for each upsert.

I also added the PRAGMA synchronous = NORMAL to the sqlite connection. This makes it so it only fsyncs on WAL checkpoints. This is recommended in WAL mode and should speed up global write lock.

### DIFF
--- a/crates/microsandbox/lib/image/mod.rs
+++ b/crates/microsandbox/lib/image/mod.rs
@@ -727,14 +727,16 @@ async fn upsert_layer_record<C: ConnectionTrait>(
         })
 }
 
-/// Attempt to satisfy `Image::persist` with a single bulk UPDATE.
+/// Attempt to satisfy `Image::persist` with a couple of bulk UPDATEs.
 ///
 /// Returns `Some(image_ref_id)` when the database is already consistent with
-/// `metadata` (i.e. the `image_ref` row exists and points to a manifest whose
-/// digest matches `metadata.manifest_digest`). In that case the only write
-/// performed is a bulk `UPDATE layer SET last_used_at` for LRU bookkeeping —
-/// the manifest, config, layer, and junction rows are content-addressed and
-/// guaranteed to be unchanged for a given manifest digest.
+/// `metadata` (i.e. the `image_ref` row exists, points to a manifest whose
+/// digest matches `metadata.manifest_digest`, and every expected `layer` row
+/// is present). In that case the only writes performed are a bulk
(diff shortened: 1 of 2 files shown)


crates/microsandbox/lib/image/mod.rs
```diff
@@ -715,3 +726,52 @@ async fn upsert_layer_record<C: ConnectionTrait>(
...
+            .layers
+            .iter()
+            .map(|layer| layer.diff_id.clone())
+            .collect();
+
+        layer_entity::Entity::update_many()
+            .col_expr(layer_entity::Column::LastUsedAt, Expr::value(now))
+            .filter(layer_entity::Column::DiffId.is_in(diff_ids))
+            .exec(db)
+            .await?;
```
toksdotdev, line 784: we should also update the image last used

crates/microsandbox/lib/image/mod.rs
```diff
@@ -715,3 +726,52 @@ async fn upsert_layer_record<C: ConnectionTrait>(
...
+/// Returns `Some(image_ref_id)` when the database is already consistent with
+/// `metadata` (i.e. the `image_ref` row exists and points to a manifest whose
+/// digest matches `metadata.manifest_digest`). In that case the only write
+/// performed is a bulk `UPDATE layer SET last_used_at` for LRU bookkeeping —
+/// the manifest, config, layer, and junction rows are content-addressed and
+/// guaranteed to be unchanged for a given manifest digest.
+///
+/// Returns `None` when the caller must fall through to the full transactional
+/// upsert (fresh DB, manifest digest changed, partially persisted state).
+async fn try_persist_fast_path(
```
toksdotdev, line 743: an extra check here to ensure all the layers in the image exists in the db would be great. 

we persist image layers [within a transaction](https://github.com/superradcompany/microsandbox/pull/665/changes#diff-ed77c0b309de3fa9a96f4f83fc686064d9971c55ba03a45f5cb0f499a43cde3aR174), sothis extra check is more of just a sanity check (just in case).

toksdotdev (approved):

PR #666, opened 2026-05-04
refactor(db): typed read/write pools with retry-on-busy

toksdotdev:
## summary

sqlite is single-writer, so when several callers in the same process try to write at once they fight for the writer lock and surface `SQLITE_BUSY` errors. this pr serialises in-process writes, makes the read/write split a property of the type system, and centralises busy handling so that contention turns into a deterministic queue rather than a flurry of retries.

each database now opens two pools wrapped as distinct types:

- `DbReadConnection` (multi-connection, concurrent reads under wal)
- `DbWriteConnection` (single connection, retry-on-busy built into its `transaction` method)

both implement `ConnectionTrait`, so existing query builders work unchanged. function signatures across the codebase declare intent: read helpers take `&DbReadConnection`, write helpers take `&DbWriteConnection`, and the few that do both take `&DbPools`.

cross-process busy contention (e.g. host vs in-vm runtime) is absorbed by an exponential-backoff retry helper. busy detection matches the `sqlx::Error` structure rather than searching the display string.

the sqlite connection builder also moves into the `microsandbox-db` crate so the host cli and the runtime apply identical pragmas (wal, busy_timeout, foreign_keys, synchronous=normal) from one place. the only user-facing tuning knob is `database.busy_timeout_secs` in `~/.microsandbox/config.json`.

## benchmark

200 concurrent boots × 3 runs, alpine, 64 mib, fresh db on each branch. the 73 vm-layer failures per run on both sides are a macos hypervisor.framework resource ceiling (memory pressure at 200 × 64 mib) and not affected by this pr, so the table normalises against the 127 boots that cleared the vm ceiling.

|                       | main (median) | branch (median) | delta   |
|-----------------------|---------------|-----------------|---------|
| succeeded             | 40 / 127      | 127 / 127       | +218%   |
| success rate          | 31%           | 100%            | +69 pts |
| wall time             | 7636 ms       | 5950 ms         | -22%    |
| throughput            | 5.6/s         | 21.3/s          | +280%   |
| `SQLITE_BUSY` errors  | 49–114/run    | 0               | gone    |

main loses 49–114 sandboxes per run to sqlite contention; the branch eliminates those failures entirely.

results are also far more deterministic on the branch: every run hits exactly 127 successful boots, while main swings between 25 and 99.

## test plan

- [x] cargo test -p microsandbox --lib
- [x] cargo check --workspace --all-targets
- [x] manual: spawn two sandboxes concurrently and confirm no SQLITE_BUSY surfaces


toksdotdev:
## addendum: linux benchmark

ran the same benchmark on a fresh gcp `n2-standard-8` (8 vcpu, 32 gib ram, ubuntu 24.04, kvm) for a cross-platform check. results are stronger than the darwin run because there's no hypervisor.framework ceiling competing with the db effect.

200 concurrent boots × 3 runs, alpine, 64 mib, fresh db on each branch:

|                       | main (median) | branch (median) | delta   |
|-----------------------|---------------|-----------------|---------|
| succeeded             | 58 / 200      | 200 / 200       | +245%   |
| success rate          | 29%           | 100%            | +71 pts |
| throughput            | 1.2 / s       | 4.4 / s         | +267%   |
| `SQLITE_BUSY` errors  | 117–143/run   | 0 (run 3 had 1) | gone    |

main loses 117–143 sandboxes per run to sqlite contention. wall time is similar on both branches (~46s) because at this concurrency the wall clock is bounded by per-vm boot time, not by db throughput. the fix moves the failure mode entirely.

toksdotdev:
## tl;dr: msb's "slowness" is nested-kvm's tax, not msb's

benchmarked on google cloud:

| metric              | nested kvm (n2-highcpu-80) | bare metal (c3-192-metal) | speedup |
|---------------------|---------------------------:|--------------------------:|--------:|
| 1-vm warmup         | 567 ms                     | 273 ms                    | 2.1×    |
| 60-conc wall        | 9246 ms                    | 391 ms                    | 24×     |
| 60-conc throughput  | 6.5/sec                    | 153/sec                   | 24×     |
| 200-conc throughput | 0–2 sandboxes/sec          | 100–130 sandboxes/sec     | ~50×    |

at 200 concurrent boots, nested kvm cliffs catastrophically: success collapses from 128/200 in run 1 to 0/200 by run 3. bare metal: **200/200 every run in ~1.5 s**.

perf showed kernel time on nested kvm dominated by ept mapping + tlb flushes; every vmexit pays a ~5–10× tax that disappears on metal.

next: a guide for nested-virt tuning. early testing suggests the available knobs (numa balancing, transparent hugepages, scheduler) only buy ~10%.

PR #672, opened 2026-05-06
chore(release): bump version to 0.4.5

toksdotdev:
bump every microsandbox crate, the node-ts and python sdks, and the typescript example pins from 0.4.4 to 0.4.5 ahead of the next release.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [main]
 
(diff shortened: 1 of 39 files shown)


appcypher (changes requested):
You should bump the version of microsandbox-mcp and its microsandbox dep, then commit.

You can also do it from within the mcp/ submodule.

After that, update the commit the submodule points to.

PR #696, opened 2026-05-10
feat(sdk/go): map KindSandboxAlreadyExists to ErrSandboxAlreadyExists

toksdotdev:
follow-up to #587 to wire the typed error so the upcoming rust-sdk replace fix flows through to go callers without another round-trip.

## changes
- added `KindSandboxAlreadyExists` to `sdk/go/internal/ffi`
- added the `wrapFFI` case mapping it to `ErrSandboxAlreadyExists`
- updated the readme error table; dropped the kind from the forward-compat list since it now has a concrete behavior
- extended the existing `kindFromFFI` test with the new case

## context
the rust-side `replace()` hang reported on #587 is being addressed in a separate pr against `crates/microsandbox`. that pr will start emitting a `sandbox_already_exists` error kind across the ffi boundary. once it lands and #587 merges, callers will get a typed `ErrSandboxAlreadyExists` they can match on with `microsandbox.IsKind`.

target is `feat/go-sdk` so the change rides along with #587 into main.

## test plan
- `go build ./...` passes
- `go test ./...` passes (kindFromFFI test extended with the new tag)


toksdotdev:
rolling this into #587 as an additional commit instead

PR #704, opened 2026-05-13
feat(build): honor MSB_HOME in prebuilt download path

nuri-yoo:
## summary

PR #669 introduced `MSB_HOME` as the override for the microsandbox home directory: utils' `resolve_home`, config, db, and `setup::install` all honor it now. The crate-level build script does not — it resolves its target from `$HOME` directly and writes the downloaded `msb` + `libkrunfw` to `$HOME/.microsandbox/{bin,lib}/` unconditionally. That means a project that points microsandbox at a non-default home with `MSB_HOME` still ends up with build-time prebuilt artifacts stranded in `$HOME/.microsandbox`, and the runtime `setup::install` then re-downloads them under `$MSB_HOME` because that is where the rest of the resolver looks.

This change switches the build script to `microsandbox_utils::resolve_home()` (already a build dependency) so build-time and runtime resolution agree. It also adds `rerun-if-env-changed=MSB_HOME` / `=HOME` so the script is re-evaluated when the resolved path changes. No behavior change when `MSB_HOME` is unset.

## test plan

- [x] `MSB_HOME=/tmp/msb_test cargo build -p microsandbox` places `msb` and `libkrunfw` under `/tmp/msb_test/{bin,lib}/`, leaves `~/.microsandbox` untouched.
- [x] Plain `cargo build -p microsandbox` keeps the existing skip-if-installed behavior against `~/.microsandbox`.
- [x] `cargo check --workspace` passes.
- [x] `cargo test --workspace` passes (59 suites, 1232 tests, 0 failed).

### DIFF
--- a/crates/microsandbox/build.rs
+++ b/crates/microsandbox/build.rs
@@ -1,4 +1,5 @@
-//! Build script — downloads prebuilt msb + libkrunfw to ~/.microsandbox/{bin,lib}/.
+//! Build script — downloads prebuilt msb + libkrunfw to `$MSB_HOME` (or
+//! `~/.microsandbox/`) under `{bin,lib}/`.
 
 use std::fs;
 use std::io::{self, Cursor, Read};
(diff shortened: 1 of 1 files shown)


toksdotdev (approved):

toksdotdev:
appreciate the fix. 🙏🏽

PR #712, opened 2026-05-13
fix(image): validate digest path components

immanuwell:
## What
Tighten OCI digest parsing to follow the descriptor grammar, and keep `to_path_safe()` safe even when a digest is built directly.

Tiny guardrail, no big refactor.

## Repro
On `main`, these pass:

```rust
assert!("sha256:../../escape".parse::<Digest>().is_ok());
assert_eq!(
    Digest::new("sha/256", "../../escape").to_path_safe(),
    "sha/256_../../escape"
);
```

That string is later used for cache filenames, so yeah, not ideal.

## Checks
- `cargo fmt --all -- --check`
- `cargo test -p microsandbox-image digest::tests`
- `cargo test -p microsandbox-image`
- `cargo clippy -p microsandbox-image -- -D warnings`

### DIFF
--- a/crates/image/lib/digest.rs
+++ b/crates/image/lib/digest.rs
@@ -44,7 +44,11 @@ impl Digest {
     ///
     /// Replaces `:` with `_` (e.g., `sha256_abc123...`).
     pub fn to_path_safe(&self) -> String {
-        format!("{}_{}", self.algorithm, self.hex)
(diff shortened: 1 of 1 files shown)


toksdotdev (approved):
thanks for the fix. much appreciated :)

PR #724, opened 2026-05-14
fix(sdk-node): accept link-local in plain-object network policies

jachris:
## Summary

`NetworkBuilder.policy(p)` rejected plain-object policies whose rules referenced a `link-local` `DestinationGroup` with `unknown variant 'link-local'`. the TS public string form is kebab-case, but the JSON path is parsed by rust serde with `rename_all = "snake_case"`, which only accepts `"link_local"`.

normalize `"link-local"` → `"link_local"` in `remapDestination` for the `group` variant only. cidr / domain / domain-suffix payloads pass through unchanged — domain names can legitimately contain hyphens.

the `NetworkPolicyBuilder` instance path (`policyFromBuilder`) was unaffected: it crosses ffi as a typed rust enum, not via serde.

## Test plan

- [x] `npm run test:unit` — 4 files / 64 tests pass
- [x] new regression test `NetworkBuilder.policy() accepts a link-local group destination` exercises the plain-object → `policyJson` path and fails without this fix

### DIFF
--- a/sdk/node-ts/src/index.ts
+++ b/sdk/node-ts/src/index.ts
@@ -228,6 +228,8 @@ hideMethod(napi.SandboxBuilder, "attachWithBuilder");
   // serializing.
   const camelToSnake = (k: string): string =>
     k.replace(/[A-Z]/g, (c) => "_" + c.toLowerCase());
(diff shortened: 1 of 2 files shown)


toksdotdev (approved):

toksdotdev:
thanks for the fix. 🙏🏽

PR #730, opened 2026-05-15
drop redundant MSB_PATH reads at python + node sdk layers

toksdotdev:
## summary

dropped redundant msb-path env-var reads at the sdk layer in python and node. both sdks were funnelling env values into rust's tier 2 via `set_sdk_msb_path`, while rust's tier 1 already honoured the same env natively at spawn time.

- python: removed `MICROSANDBOX_MSB_PATH` read in `_runtime.py`
- node: removed `process.env.MSB_PATH` read in `resolve-binary.ts` + `napi.ts`

aligns both sdks with the go sdk pattern.

## breaking change

python users with `MICROSANDBOX_MSB_PATH=…` in their scripts must migrate to `MSB_PATH=…`. node users see no behaviour change. the `npx microsandbox` cli shim is unchanged.

## test plan

- [x] python: lint, unit tests, end-to-end sandbox with + without env vars
- [x] node: typecheck, build, end-to-end sandbox with + without env vars
- [x] confirmed `MICROSANDBOX_MSB_PATH` is silently ignored by python after the change

### DIFF
--- a/sdk/node-ts/src/internal/napi.ts
+++ b/sdk/node-ts/src/internal/napi.ts
@@ -1,10 +1,11 @@
 import { createRequire } from "node:module";
 import { msbPath } from "./resolve-binary.js";
 
-// Resolve the runtime binary once. User-provided MSB_PATH still wins, but the
-// SDK passes the resolved path to native code explicitly instead of relying on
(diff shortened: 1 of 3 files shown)


appcypher (approved):
LGTM

PR #733, opened 2026-05-15
feat(cli): change start cmd to take variadic args

### LINKED ISSUE #692: Variadic arguments for msb start/stop
`msb remove` supports variadic args so you can do something like this:

```
msb remove sandbox-1 sandbox-2 sandbox
```

However msb start/stop only support one arg:

```
msb stop sandbox-1
msb start sandbox-1
```

cedws:
Closes #692.


appcypher (approved):

crates/cli/lib/commands/start.rs
```diff
@@ -26,23 +27,33 @@ pub struct StartArgs {
...
+            Ok(sandbox) => {
+                sandbox.detach().await;
+                spinner.finish_success("Started");
+            }
+            Err(e) => {
+                spinner.finish_clear();
+                if !args.quiet {
+                    ui::error(&format!("{e}"));
+                }
+                failed = true;
```
toksdotdev, line 49: instead of an outright failed, we should show a list of the sandboxes, and their corresponding status

appcypher: @toksdotdev it does that already. ui::error(...) shows the status of each failed sandbox. 

setting `failed` here allows us return error code 1 to show that at least one sandbox failed to start properly.

there is similar logic in ../stop.rs

PR #734, opened 2026-05-15
docs(sdk): document blocking/nonblocking nature of exec funcs

cedws:
Discussed briefly with Toks. It's useful to have the docstrings disambiguate blocking vs nonblocking calls in cases like this.


toksdotdev (approved):
lgtm!

PR #749, opened 2026-05-16
fix(cli logs): emit JSON Lines in follow mode too

dijdzv:
`msb logs --follow --json` emits valid JSON Lines for the initial dump (via `render_entries`) but silently degrades to plain text once the follow loop takes over, because `render_one` was ignoring `--json`. Downstream consumers parsing each line as JSON break the moment new records start arriving through the follow path.

This patch extracts a `write_entry_json(entry, &mut out)` helper and routes both render paths through it, so the field set (`{t, s, d, id, e}`, same shape `render_entries` already emitted) cannot drift between the two callers.

### Verification

Repro: `msb logs <sandbox> --follow --json | jq .`

| Phase | Before | After |
|---|---|---|
| Initial dump | JSON ✓ | JSON ✓ |
| Follow records | plain text ✗ | JSON ✓ |
| `jq` consumer | parse error mid-stream | parses every line |

### DIFF
--- a/crates/cli/lib/commands/logs.rs
+++ b/crates/cli/lib/commands/logs.rs
@@ -490,22 +490,26 @@ fn parse_entry_time(t: &str) -> Option<DateTime<Utc>> {
 // Functions: Helpers — rendering
 //--------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


toksdotdev (approved):

toksdotdev:
thanks for the fix.

PR #759, opened 2026-05-19
fix(sdk/python): argument passing in `Sandbox.exec(str, dict)` case

vikanezrimaya:
`args` went unused when passing an options dict in the Python SDK.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -511,6 +511,73 @@ jobs:
         run: bunx --bun vitest run
 
   # ---------------------------------------------------------------------------
+  # Python SDK integration tests (requires KVM)
+  #
+  # Downloads the pre-built msb + libkrunfw from the `check` job, then builds
+  # the Python extension locally and runs the integration suite against those
+  # fresh runtime binaries.
+  # ---------------------------------------------------------------------------
+  python-sdk-test:
+    name: Python SDK Tests
+    needs: [check, changes]
+    if: needs.changes.outputs.code == 'true'
+    runs-on: self-hosted-ubuntu-2404-x64
+    steps:
+      - name: Clean workspace
+        run: |
+          rm -rf "${{ github.workspace }}"/{build}
+          rm -rf ~/.microsandbox
+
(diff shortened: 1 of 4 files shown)


vikanezrimaya:
I'm surprised there wasn't a test that caught this before. Looks like the Python SDK's Rust code is not covered by any tests at all? Should I even *attempt* to write a test?

It looks like the Rust-side tests usually fetch `mirror.gcr.io/library/alpine` as the image, and these tests run on self-hosted KVM-enabled runners. Python tests currently do not seem to do that (run on default runners, as far as I understand? I don't use GitHub and its CI much.)

My opinion is this will require changes to the CI setup to add another step to run Python tests requiring KVM on a KVM-enabled machine (similar to Node smoke tests?).

Anyway, here's how I caught this: (in an asyncio repl)

```python
import asyncio
import microsandbox as msb

haneul = await (await msb.Sandbox.get("haneul")).connect()
assert (await haneul.exec("echo", {"args": ["hello"]})).stdout_text == "hello\n"
```

toksdotdev:
hey, thanks for fix. it's on our roadmap to step up proper integration test for python like the other SDKs.

i'll get up and running the python CI setup shortly, and rebase on this. yes, test for this would be great!

toksdotdev:
@vikanezrimaya no idea why github closed the pr after rebasing on latest main. can you re-open the pr when you've got some time?

vikanezrimaya:
@toksdotdev I think you might've done a force-push instead of a rebase and nuked my commit. With no new commits, GitHub simply closed the PR. No worries, I think can rebase on top of main and try to push again.

vikanezrimaya:
@toksdotdev oh, no, it's much worse. You seem to have accidentally folded my commit into 9c43918298857a33b3b8d8b28c139340969e7495! So… I suppose you effectively merged it. (But you lost my signature and attribution in the process.)

I think fixing this will require rewriting history and coordinating effort due to commit signing. I'm not sure whether it's a good idea considering the "rewriting history" part.

PR #771, opened 2026-05-20
Passthrough secret violation type

wellsbunk5:
Adds `Passthrough` as a valid secret violation policy. When a matching request contains a secret placeholder, the request is not blocked. Instead, the placeholder is forwarded unchanged and is not substituted with the real secret value.

Passthrough is now configured through the secret violation policy itself. `ViolationAction::Passthrough` carries the host patterns that may receive placeholders, along with a fallback action for non-matching hosts.

This also adds a dedicated violation action builder for both global and per-secret configuration. `NetworkBuilder::on_secret_violation(...)` and `SecretBuilder::on_violation(...)` now accept a builder closure, so callers can define the fallback behavior and passthrough hosts in one place. These passthrough settings only allow placeholders through; they do not authorize real secret substitution.


crates/network/lib/builder.rs
```diff
@@ -178,6 +179,7 @@ impl NetworkBuilder {
             value: value.into(),
             placeholder: placeholder.into(),
             allowed_hosts: vec![HostPattern::Exact(allowed_host.into())],
+            passthrough_hosts: Vec::new(),
```
toksdotdev: hmm. instead of a dedicated passthrough hosts field, we can have a `on_violation: ViolationAction` field here, and then introduce a new variant on `ViolationAction::Passthrough(Vec<HostPattern>)`.

same applies to other places across the code.

crates/cli/lib/commands/common.rs
```diff
@@ -967,8 +967,9 @@ fn parse_violation_action(
         Some("block") => Ok(Some(ViolationAction::Block)),
         Some("block-and-log") => Ok(Some(ViolationAction::BlockAndLog)),
         Some("block-and-terminate") => Ok(Some(ViolationAction::BlockAndTerminate)),
+        Some("passthrough") => Ok(Some(ViolationAction::Passthrough)),
```
toksdotdev: based on the other comment, this would then be mapped to `::Passthrough([HostPattern::Any])`.

representing passthrough for multiple hosts is going to be a complex syntax (especially for secrets), so no need to support that today. there'll eventually be support for specifying a yaml config to the cli for such nested fields.

toksdotdev (approved):

toksdotdev:
thanks a lot, @wellsbunk5 🫡 🙌🏽

PR #773, opened 2026-05-22
docs(sdk/go): add cpu/memory defaults to docstrings

cedws:
Since the defaults are in the core Rust library, these docstrings can become easily desynced unfortunately, but it's useful to not have to go digging in the MSB code to find out what the defaults are.


toksdotdev (approved):
lgtm! thanks.

PR #780, opened 2026-05-23
Fix published-port stalls by waking smoltcp after host accepts

billroy:
## Problem

This proposed fix resolves a serious user-facing problem where sites which load instantly locally were taking 15 to 20 seconds to load when run from a sandbox.

Host-to-guest published-port traffic can stall for seconds even when the guest service is ready and responds immediately. In a local reproduction, requests from the host to a published port connected and sent successfully, but then waited in TTFB plateaus of several seconds, sometimes around 10 seconds. The same service, when requested from inside the guest, responded in low milliseconds.

A minimal diagnostic HTTP server reproduced the stall without the application stack involved, which isolated the issue to the Microsandbox published-port forwarding path.

The root cause is that the host-side TCP listener accepts a connection and sends it into `PortPublisher`'s inbound channel, but it does not wake the smoltcp poll loop. If the poll loop is asleep in `poll(2)`, the accepted connection can sit in that channel until an unrelated wake or timer causes the loop to run again.

## Proposed Fix

Wake `shared.proxy_wake` immediately after a published-port listener successfully queues an accepted host connection.

The relay task already wakes this same poll loop after reading host socket data, but that happens too late for this case: the relay task is only spawned after the poll loop drains the inbound connection queue and creates the smoltcp socket. The missing wake is therefore on the accept/queue edge.

## Testing

- `cargo fmt --check`
- `cargo check -p microsandbox-network`
- `cargo test -p microsandbox-network queue_inbound_connection_wakes_poll_loop`
- `cargo test -p microsandbox-network --lib`
- `cargo clippy -p microsandbox-network -- -D warnings`

## Local Validation

With the patch applied to a local `msb`, the diagnostic server no longer showed the multi-second TTFB plateaus. The affected application also returned to low-ms load timings through the same published-port path.

### DIFF
--- a/crates/network/lib/publisher.rs
+++ b/crates/network/lib/publisher.rs
@@ -353,12 +353,25 @@ async fn tcp_listener_task(
         }
 
         let conn = InboundConnection { stream, guest_port };
-        if inbound_tx.send(conn).await.is_err() {
+        if !queue_inbound_connection(&inbound_tx, conn, &shared).await {
             break; // Publisher dropped.
         }
     }
 
     Ok(())
+}
+
+async fn queue_inbound_connection<T>(
+    inbound_tx: &mpsc::Sender<T>,
+    conn: T,
+    shared: &SharedState,
+) -> bool {
+    if inbound_tx.send(conn).await.is_err() {
+        return false;
+    }
+
+    shared.proxy_wake.wake();
+    true
 }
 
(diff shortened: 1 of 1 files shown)


billroy:
For the debugging record, here is the Bullpen post-mortem for this issue: https://github.com/billroy/bullpen/blob/main/docs/microsandbox-http-performance-issue.md

PR #783, opened 2026-05-24
fix(secrets): keep request body as raw bytes through substitute()

jachris:
## Summary
`SecretsHandler::substitute()` rebuilt the request body via `String::from_utf8_lossy`, which rewrites every non-UTF-8 byte to U+FFFD. Binary request bodies — anything with `Content-Encoding: gzip`, any compressed or otherwise non-UTF-8 upload — round-tripped corrupted.

## Changes
- Carry the body through as `&[u8]`. Body placeholder substitution goes via the existing `contains_bytes` plus a new `replace_bytes` helper, so `inject_body=true` secrets still work while binary payloads pass through exactly. Headers stay text and are still lossy-decoded.

## Related pre-existing issue
While I was in this code I noticed a placeholder leak worth surfacing. `substitute()` runs once per rustls `read(buf)`, so a single request can split between a headers read and a body read. Any chunk without `\r\n\r\n` is parsed as headers, and the header substitution path runs on its bytes. For a secret configured `inject_headers=true, inject_body=false`, a literal placeholder in the body gets replaced with the real secret value and forwarded to the server — leaking the secret into a payload the user opted out of injecting into. This is present on `main` today and unchanged by this PR; #784 will address it.

## Test plan
- [x] `cargo test -p microsandbox-network --lib secrets::` — the new regression test fails on `main`, passes after the fix.

### DIFF
--- a/crates/network/lib/secrets/handler.rs
+++ b/crates/network/lib/secrets/handler.rs
@@ -947,4 +947,34 @@ mod tests {
         ));
         assert!(!json_escaped_contains(b"KEY", b"$KEY"));
     }
+
+    #[test]
+    fn body_preserves_non_utf8_bytes() {
+        // substitute() operates on the raw plaintext byte stream. The body
+        // portion must round-trip byte-for-byte regardless of content;
+        // request bodies are arbitrary application data (compressed payloads,
+        // binary uploads, …) and the function has no license to transform
+        // bytes outside its declared substitution scopes.
+        //
+        // Uses default injection scopes (headers + basic_auth, no body) so
+        // the body is out of scope for substitution but the eligible secret
+        // still forces the function down its full processing path.
+        let config = make_config(vec![make_secret("$KEY", "real-secret", "example.com")]);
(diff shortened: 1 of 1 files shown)


jachris:
@toksdotdev I just noticed that this is pretty much a duplicate of #768, which didn't exist when I already implemented this fix locally. Sorry, happy to close this if yours lands. And just fyi, neither PR is enough to fully address the issue linked in your PR. I also had to stack #784 on top of it. Together, this makes `git`-in-sandbox work reliably in my tests.

toksdotdev:
hey @jachris no worries. i'll aim to get in #768 asap, and will review the PRs you put up. really appreciate the fix 🙏🏽

toksdotdev:
@jachris mind rebasing when you've got some time. 😃 the changes here does look similar, but if i could see the diff that's be helpful (just in case i missed an edge case).

jachris:
I checked: the diff would either be empty or a revert of your PR, depending on which side your pick. So you didn't miss anything. I'll rebase the other PR now on top of your changes, so we can close this.

PR #784, opened 2026-05-24
fix(secrets): handle body-continuation chunks across rustls reads

jachris:
## Why

The TLS proxy calls `substitute()` once per rustls plaintext chunk, and two facts about HTTP/1.1 collide with that model:

1. **Multiple requests per connection.** Keep-alive is the default; one `SecretsHandler` instance sees a sequence of requests over a single connection. Some clients pipeline, sending several requests back-to-back without waiting for responses.
2. **TLS chunks don't align with HTTP framing.** Even a single request commonly arrives as `(headers + start of body)` then `(more body)` then `(rest of body)`. Continuation chunks don't contain `\r\n\r\n`. And a single chunk can also span request boundaries: the tail of one request's body plus the head of the next request's headers.

#768 made `substitute()` handle headers vs body correctly within a single chunk for a single request. It has no concept of "this chunk continues the previous request's body" or "this chunk contains the start of the next request." Three bugs fall out, all load-bearing:

- **Silent body corruption.** When a body-continuation chunk arrives, body bytes go through `String::from_utf8_lossy` to be scanned for header substitution. Any non-UTF-8 byte (gzip uploads, compressed payloads, anything binary) becomes U+FFFD on the wire.
- **Header-only secrets leak into bodies.** A secret configured `inject_headers=true, inject_body=false` runs `substitute_in_headers` on bytes that are actually body. A literal `$KEY` anywhere in the body is replaced with the real secret value and forwarded, which is the opposite of what the user asked for.
- **Pipelined next-request headers are forwarded raw.** When a chunk contains the end of request 1 plus the start of request 2, the spillover bytes for request 2 are emitted verbatim. Their headers never go through substitution or base-auth violation detection.

## Changes

Two pieces.

**HTTP framing state on `SecretsHandler`:**

- `HttpState::AwaitingHeaders`: scanning for the next `\r\n\r\n`.
- `HttpState::InBody { remaining: Option<usize> }`: inside a body. `Some(n)` counts down per Content-Length. `None` is unknown framing (chunked or connection-close); stay in this state until the connection ends.

When `substitute()` enters in `InBody`, dispatch to `substitute_body_chunk`: violation scan, decrement `remaining`, return `Cow::Borrowed(data)`. Never run `from_utf8_lossy`, never run header substitution.

**Pipelining-aware splitting:**

Both `substitute()` and `substitute_body_chunk` now split the incoming chunk into `(this_request, spillover)` and recurse on the spillover with state advanced to `AwaitingHeaders`. The shared helper `append_pipelined_spillover` keeps the zero-copy `Cow::Borrowed(parent)` fast path when both halves pass through unchanged; otherwise it allocates and concatenates. `next_state_after_headers` now returns `(HttpState, body_in_request)` so the caller can split off spillover bytes when `body_in_chunk > Content-Length`.

## Failing tests on `main`

Six new tests fail on `main` and pass on this branch (three are body-corruption / leak regressions, three are pipelining regressions).

<details>
<summary><code>body_in_separate_chunk_preserves_non_utf8_bytes</code>: headers in chunk 1, body in chunk 2 with <code>$KEY</code> between non-UTF-8 bytes. Asserts the body passes through verbatim.</summary>

```
assertion `left == right` failed
  left: [0, 239, 191, 189, 239, 191, 189, 239, 191, 189, 239, 191, 189, 114, 101, 97, 108, 45, 115, 101, 99, 114, 101, 116, 239, 191, 189, 239, 191, 189, 239, 191, 189, 239, 191, 189]
 right: [0, 128, 192, 255, 254, 36, 75, 69, 89, 129, 193, 238, 239]
```

Each `EF BF BD` is U+FFFD (one non-UTF-8 byte lossy-replaced). The `$KEY` token (36, 75, 69, 89) has been replaced with `real-secret` (114, 101, 97, 108, 45, 115, 101, 99, 114, 101, 116). Both the corruption and the leak in one chunk.

</details>

<details>
<summary><code>body_split_across_two_chunks_round_trips</code>: chunk 1 carries headers + first slice of body; chunk 2 carries the rest with <code>$KEY</code> in it. Asserts both halves pass verbatim.</summary>

```
assertion `left == right` failed
  left: [239, 191, 189, 239, 191, 189, 114, 101, 97, 108, 45, 115, 101, 99, 114, 101, 116, 239, 191, 189, 239, 191, 189, 239, 191, 189, 239, 191, 189]
 right: [253, 252, 36, 75, 69, 89, 129, 193, 238, 239]
```

Same pattern: surrounding bytes mangled to U+FFFD, `$KEY` replaced with `real-secret`.

</details>

<details>
<summary><code>header_only_secret_does_not_leak_into_body_continuation_chunk</code>: headers in chunk 1, ASCII body containing a literal <code>$KEY</code> in chunk 2. Secret is <code>inject_headers=true, inject_body=false</code>. Asserts the body's <code>$KEY</code> is forwarded verbatim.</summary>

```
assertion `left == right` failed
  left: [112, 114, 101, 102, 105, 120, 58, 114, 101, 97, 108, 45, 115, 101, 99, 114, 101, 116, 58, 109, 111, 114, 101, 45, 112, 97, 100, 100, 105, 110, 103]
 right: [112, 114, 101, 102, 105, 120, 58, 36, 75, 69, 89, 58, 109, 111, 114, 101, 45, 112, 97, 100, 100, 105, 110, 103]
```

Decoded: `"prefix:real-secret:more-padding"` (actual) vs `"prefix:$KEY:more-padding"` (expected). The leak with no binary bytes around it, for clarity.

</details>

<details>
<summary><code>pipelined_request_in_body_continuation_chunk_is_substituted</code>: chunk 1 is the headers + start of body of request 1; chunk 2 carries the last body byte plus a complete pipelined request 2 with a substitutable placeholder in a header.</summary>

```
assertion `left == right` failed
  left:  [101, 71, 69, 84, ..., 65, 117, 116, 104, 58, 32, 36, 75, 69, 89, 13, 10, 13, 10]
  right: [101, 71, 69, 84, ..., 65, 117, 116, 104, 58, 32, 114, 101, 97, 108, 45, 115, 101, 99, 114, 101, 116, 13, 10, 13, 10]
```

`$KEY` was forwarded as part of "request 1's body" instead of getting parsed as request 2's header and substituted to `real-secret`.

</details>

<details>
<summary><code>pipelined_request_in_same_chunk_as_headers_is_substituted</code>: a single chunk carries request 1's headers + complete body + the start of request 2.</summary>

```
assertion `left == right` failed
  left:  [..., 97, 98, 99, 100, 101, 71, 69, 84, ..., 65, 117, 116, 104, 58, 32, 36, 75, 69, 89, 13, 10, 13, 10]
  right: [..., 97, 98, 99, 100, 101, 71, 69, 84, ..., 65, 117, 116, 104, 58, 32, 114, 101, 97, 108, 45, 115, 101, 99, 114, 101, 116, 13, 10, 13, 10]
```

Same failure shape, from the headers-path side: request 2's bytes were treated as request 1's body, so substitution never ran on them.

</details>

<details>
<summary><code>violation_in_pipelined_next_request_basic_auth_is_detected</code>: pipelined request 2 carries <code>Authorization: Basic YWRtaW46JEtFWQ==</code> (base64 of <code>admin:$KEY</code>). The placeholder is for a disallowed host, so the call must return <code>ViolationAction::Block</code>.</summary>

```
called `Result::unwrap_err()` on an `Ok` value: [102, 111, 111, ..., 65, 117, 116, 104, 111, 114, 105, 122, 97, 116, 105, 111, 110, 58, 32, 66, 97, 115, 105, 99, 32, 89, 87, 82, 116, 97, 87, 52, 54, 74, 69, 116, 70, 87, 81, 61, 61, 13, 10, 13, 10]
```

`$KEY` is hidden inside base64, so the body-path byte scan can't see it. Only the recursive header pass that decodes Basic credentials catches the violation. Currently it returns `Ok` and forwards the credentials.

</details>

## Other tests

- `framing_state_resets_after_request_completes`: second request on a keep-alive connection parses correctly.
- `violation_detected_in_body_continuation_chunk`: byte-level placeholder in a body chunk still triggers blocking.
- `three_pipelined_requests_in_one_chunk_all_substitute`: recursion nests twice and substitutes each of three pipelined requests' headers.
- `pipelined_spillover_without_substitution_stays_zero_copy`: a chunk with a pipelined boundary but no eligible secret returns `Cow::Borrowed` (no allocation).
- `pipelined_get_without_content_length_recurses_into_next_request`: a chain of GETs without `Content-Length` is treated as zero-body per RFC 9112 §6.3 case 6; trailing bytes are the next request's headers, not body of the first.
- `substitution_resumes_after_chunked_request_body_terminator`: after the chunked end marker `0\r\n\r\n`, the next request on the same keep-alive connection is parsed and substituted normally instead of being swallowed as body.

## Known limitations

- **Multi-chunk body substitution.** A `FIXME` is left at the `Content-Length` rewrite: when `inject_body=true` and substitution changes body length and the body spans chunks, only chunk 1's body is substituted but Content-Length reflects chunk 1's new size. Continuation bytes then flow past a stale CL. Pre-existing limitation, not introduced here. The proper fix is to rewrite the request to `Transfer-Encoding: chunked` so the per-chunk body length is self-framed; out of scope for this PR.
- **Header boundary split across TLS reads.** `find_header_boundary` only searches the current chunk. If `\r\n\r\n` straddles a rustls read boundary (e.g. headers ending in `\r\n`, next read starting with `\r\n`), both chunks see no boundary, framing state never advances, and the second chunk's body bytes are run through the header substitution path - reintroducing the same binary-corruption / header-only-secret-in-body shape this PR is meant to fix for that read split. Closing it properly requires buffering header bytes until the boundary lands; the chunked-terminator-split case has the same shape. Separate PR.

## Test plan

- [x] `cargo test -p microsandbox-network --lib secrets::` (52 tests pass on this branch; the eight pipelining/framing tests above fail on `main`).

### DIFF
--- a/crates/network/lib/secrets/handler.rs
+++ b/crates/network/lib/secrets/handler.rs
@@ -44,7 +44,7 @@ enum HttpState {
     AwaitingHeaders,
     /// Inside a request body. `remaining` is the number of body bytes left
     /// per Content-Length; `None` means unknown framing (chunked or
-    /// connection-close) — stay in this state until the connection ends.
+    /// connection-close).
     InBody { remaining: Option<usize> },
 }
 
@@ -253,9 +253,7 @@ impl SecretsHandler {
     /// disallowed host.
     pub fn substitute<'a>(&mut self, data: &'a [u8]) -> Result<Cow<'a, [u8]>, ViolationAction> {
         // Body-continuation chunk: previous chunk(s) already contained the
-        // request line and headers. The remainder of the body must pass
-        // through as raw bytes — never run `from_utf8_lossy` on it, which
-        // would corrupt binary payloads.
+        // request line and headers.
         if let HttpState::InBody { remaining } = self.http_state {
             return self.substitute_body_chunk(data, remaining);
         }
@@ -263,14 +261,33 @@ impl SecretsHandler {
         // Split raw bytes at the header boundary BEFORE converting to owned strings.
         // This avoids position shifts from from_utf8_lossy replacement chars.
         let boundary = find_header_boundary(data);
-        let (header_bytes, body_bytes) = match boundary {
+        let (header_bytes, after_headers) = match boundary {
             Some(pos) => (&data[..pos], &data[pos..]),
             None => (data, &[] as &[u8]),
         };
 
+        // A single chunk may carry headers + body + the start of the next
+        // pipelined request. Compute how many post-boundary bytes belong to
+        // THIS request; the rest is spillover that gets its own recursive
+        // pass through `substitute()` so its headers are substituted and
+        // its violations are detected.
+        let (body_bytes, spillover) = if boundary.is_some() {
+            let (next_state, body_in_request) = next_state_after_headers(
+                String::from_utf8_lossy(header_bytes).as_ref(),
+                after_headers.len(),
+            );
+            self.http_state = next_state;
+            after_headers.split_at(body_in_request)
+        } else {
+            (after_headers, &[] as &[u8])
+        };
+
+        // Everything from `data` belonging to this request, headers and body.
+        let this_request = &data[..header_bytes.len() + body_bytes.len()];
+
         // Check for disallowed placeholders before forwarding or substituting data.
-        if let Some(action) =
-            self.detect_blocking_action(data, String::from_utf8_lossy(header_bytes).as_ref())
+        if let Some(action) = self
+            .detect_blocking_action(this_request, String::from_utf8_lossy(header_bytes).as_ref())
         {
             match action {
                 BlockingAction::Block => return Err(action.into_violation_action()),
@@ -280,27 +297,18 @@ impl SecretsHandler {
                 }
                 BlockingAction::BlockAndTerminate => {
                     tracing::error!(
-                        "secret violation: placeholder detected for disallowed host — terminating"
+                        "secret violation: placeholder detected for disallowed host - terminating"
                     );
                     return Err(action.into_violation_action());
                 }
             }
         }
-        self.update_tail(data);
-
-        // Update framing state for the next chunk. Done regardless of whether
-        // we end up modifying the bytes, so subsequent body-continuation
-        // chunks take the body-only path above.
-        if boundary.is_some() {
-            self.http_state = next_state_after_headers(
-                String::from_utf8_lossy(header_bytes).as_ref(),
-                body_bytes.len(),
-            );
-        }
+        self.update_tail(this_request);
 
         if self.eligible_for_substitution.is_empty() {
-            // No substitution needed. Return borrowed slice (zero-copy).
-            return Ok(Cow::Borrowed(data));
+            // No substitution needed; pass this request through and let the
+            // recursive call handle the spillover (if any).
+            return self.append_pipelined_spillover(data, this_request, spillover);
         }
 
         // Start with borrowed bytes; allocate only when a substitution is needed.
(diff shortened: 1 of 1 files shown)


jachris:
Hey @toksdotdev. I rebased this on top of main and polished it a bit. Happy to make further adjustments as needed. Let me know what you think, thanks.

toksdotdev:
awesome! thanks. will take a look today 😃

appcypher (approved):

appcypher:
Btw, i'm expanding on this in https://github.com/superradcompany/microsandbox/pull/836

PR #801, opened 2026-05-27
feat(sandbox): atomically create per-sandbox named volumes via MountBuilder::create_named

nuri-yoo:
## Summary

Adds `m.create_named(name)` and `m.create_named_with(name, |v| ...)` to the existing `MountBuilder`, following @appcypher's suggestion to express the behavior through the mount builder instead of a parallel `auto_volume*` surface. The volume row is inserted in the same DB transaction as the sandbox row, so a concurrent `Volume::list` cannot observe the volume before the owning sandbox exists.

```rust
SandboxBuilder::new("my-sandbox")
    .image("alpine")
    .volume("/tmp", |m| m.create_named("my-sandbox-tmp"))
    .volume("/scratch", |m| m
        .create_named_with("my-sandbox-scratch", |v| v
            .quota(1.gib())
            .label("owner", "my-team"))
        .readonly())
    .create_detached().await?;
```

## Motivation

Callers who want a per-sandbox dedicated volume call `Volume::create` and then list it in `SandboxConfig::mounts`. Those are two independent DB writes; another process scanning `Volume::list` between them sees an orphan-shaped row. Anyone scanning `Volume::list` for owner-less volumes (a startup sweep for orphan cleanup, a quota auditor, a metrics collector) hits the same window — no way to tell an in-flight `Volume::create` apart from a real orphan without a `created_at`-keyed grace window. Closing the window at the source removes the need for that grace window entirely.

## API

* `MountBuilder::create_named(name)` — sibling of `named()`. Same mount semantics, plus atomic provisioning of the named volume.
* `MountBuilder::create_named_with(name, |v| ...)` — `VolumeBuilder` sub-builder for `quota`, `label`, `name` override. Mount-side options (`readonly`, `noexec`, ...) stay on `MountBuilder`.
* Existing `m.named(...)` continues to mean "mount existing named volume" and is unchanged.

## Implementation

* `Volume::create_in_transaction(&txn, &VolumeConfig)` — new `pub(crate)` helper that inserts the volume row using the caller's connection. Shares validation, label serialisation, and the `kind` / `capacity_bytes` / `disk_format` / `disk_fstype` columns with `Volume::create`.
* `Volume::materialise_for(&VolumeConfig)` — provisions the on-disk artifact (directory or `disk.raw`) via the same tempdir → rename pattern as `Volume::create`. Disk volumes get ext4 formatting for free.
* `MountBuilder::create_intent: Option<AutoVolumeIntent>` — `SandboxBuilder::volume()` calls `take_create_intent()` after the closure runs and pushes any intent to `config.auto_volumes`.
* `SandboxConfig::auto_volumes: Vec<AutoVolumeIntent>` — new `#[serde(skip)] pub(crate)` transient field.
* `insert_sandbox_record` inserts every auto-volume row inside the same transaction as the sandbox row. Any error rolls both back atomically.
* `Sandbox::create_with_mode` materialises on-disk artifacts after the transaction commits. Directory failure or sandbox-spawn failure removes the volume rows and rolls back the artifacts.

## Tests

* `test_insert_sandbox_record_atomic_with_auto_volumes` — sandbox row and labelled volume row both visible after one transaction.
* `test_insert_sandbox_record_rolls_back_when_volume_name_collides` — pre-seeded duplicate volume name fails the transaction; the sandbox row never lands.
* `test_create_named_registers_intent_and_mount` — `create_named` registers the intent and emits a matching `VolumeMount::Named`.
* `test_create_named_with_carries_quota_and_labels` — sub-builder carries quota and labels, and outer `readonly()` reaches the mount.
* `test_create_named_with_name_override` — `v.name(...)` wins over the initial argument.

## Test plan

- [x] `cargo build -p microsandbox`
- [x] `cargo test -p microsandbox --lib` — 269 passed, 0 failed (264 pre-existing + 5 new related)
- [x] `cargo fmt --check -p microsandbox`

## Reference

- brekkylab/ailoy#416

### DIFF
--- a/crates/microsandbox/lib/sandbox/builder.rs
+++ b/crates/microsandbox/lib/sandbox/builder.rs
@@ -672,14 +672,28 @@ impl SandboxBuilder {
     /// .volume("/data", |m| m.bind("/host/data"))
     /// .volume("/config", |m| m.bind("/host/config").readonly())
     /// .volume("/cache", |m| m.named("my-cache"))
-    /// .volume("/tmp", |m| m.tmpfs().size(100))
+    /// .volume("/tmp", |m| m.create_named("my-sandbox-tmp"))
+    /// .volume("/scratch", |m| m
+    ///     .create_named_with("my-sandbox-scratch", |v| v.quota(1024).label("owner", "team"))
+    ///     .readonly())
+    /// .volume("/tmpfs", |m| m.tmpfs().size(100))
     /// ```
     pub fn volume(
         mut self,
         guest_path: impl Into<String>,
         f: impl FnOnce(MountBuilder) -> MountBuilder,
     ) -> Self {
-        match f(MountBuilder::new(guest_path)).build() {
+        let mut mb = f(MountBuilder::new(guest_path));
+        if let Some(intent) = mb.take_create_intent() {
+            if let Err(e) = crate::volume::validate_volume_name(&intent.name)
+                && self.build_error.is_none()
+            {
+                self.build_error = Some(e);
+                return self;
+            }
+            self.config.auto_volumes.push(intent);
+        }
+        match mb.build() {
             Ok(mount) => self.config.mounts.push(mount),
             Err(e) => {
                 if self.build_error.is_none() {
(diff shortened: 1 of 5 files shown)


appcypher:
I'm not sure how to proceed with this PR. On the one hand, it's a useful feature to have. On the other, it diverges quite a bit from the current API.

I'm trying to come up with a fine middleground.

appcypher:
Could we keep the behavior but express it through the existing mount builder instead?

Example:
```rs
    .volume("/tmp", |m| m.create_named("my-sandbox-tmp"))
    .volume("/scratch", |m| {
        m.create_named_with("my-sandbox-scratch", |v| {
            v.quota(1.gib()).label("owner", "my-team")
        }).readonly()
    })
```

Then:
- `m.named("foo")` continues to mean “mount existing named volume”
- `m.create_named("foo")` means “create this named volume atomically with the sandbox, then mount it”
- volume metadata stays in a VolumeBuilder-like API
- mount options stay on MountBuilder

That would preserve the atomic transaction work from this PR while fitting the existing API more closely.

nuri-yoo:
> Could we keep the behavior but express it through the existing mount builder instead?
>
> Example:
> ```rs
>     .volume("/tmp", |m| m.create_named("my-sandbox-tmp"))
>     .volume("/scratch", |m| {
>         m.create_named_with("my-sandbox-scratch", |v| {
>             v.quota(1.gib()).label("owner", "my-team")
>         }).readonly()
>     })
> ```
>
> Then:
> - `m.named("foo")` continues to mean "mount existing named volume"
> - `m.create_named("foo")` means "create this named volume atomically with the sandbox, then mount it"
> - volume metadata stays in a VolumeBuilder-like API
> - mount options stay on MountBuilder
>
> That would preserve the atomic transaction work from this PR while fitting the existing API more closely.

Reworked along these lines. Summary of the changes vs the previous revision:

- `MountBuilder::create_named(name)` and `create_named_with(name, |v| ...)` replace the `auto_volume*` surface. `m.named` is unchanged.
- `VolumeBuilder` sub-builder carries `quota`, `label`, `name` override. `readonly` and other mount options stay on `MountBuilder`, as suggested.
- Atomic transaction work is preserved: `Volume::create_in_transaction` inserts the volume row inside the sandbox transaction; `Volume::materialise_for` provisions the on-disk artifact via the same `tempdir → rename` pattern as `Volume::create`, so disk volumes (ext4-formatted `disk.raw`) work as well.

One small deviation: I dropped the implicit `{sandbox_name}-{guest_slug}` default name and made the volume name a required argument on both methods, since the closure form makes the explicit name read more naturally and there's no `auto_volume(guest_path)` zero-arg variant to mirror. Happy to add a default back if you'd rather keep that ergonomic.

Rebased onto current `main`; 269/269 lib tests pass.

appcypher (approved):

PR #822, opened 2026-05-27
Enable nested virtualization for Linux x86_64 VMs

enricoschaaf:
## Summary

- Enable libkrun nested virtualization for Linux x86_64 VMs by calling `nested_virt(true)` in the runtime machine builder.
- Keep the existing split irqchip setup on the same platform-specific path.

## Notes

This exposes nested virtualization support from the outer VMM configuration. A fully usable nested KVM setup still depends on the host having nested virtualization enabled and the guest firmware/kernel including KVM support.

## Testing

Not run, per request.


enricoschaaf:
Closing so this can be recreated from the renamed branch without the green-got prefix.

PR #823, opened 2026-05-27
Enable nested virtualization for runtime VMs

enricoschaaf:
## Summary

- Enable libkrun nested virtualization for Linux x86_64 runtime VMs.
- Keep nested virtualization scoped to the Intel/AMD path where Microsandbox is currently deployed.
- Keep the existing split irqchip setup on the Linux x86_64-specific path.

## Notes

This exposes nested virtualization support from the outer VMM configuration. A fully usable nested KVM setup still depends on the host/hypervisor supporting nested virtualization and the guest firmware/kernel including the relevant KVM support. The matching firmware config change is in superradcompany/libkrunfw#7.

## Testing

Not run in this checkout. The same runtime setting was validated in a Green Got sandbox-host build by creating a sandbox and confirming `/dev/kvm`, VMX CPU flags, and `KVM_GET_API_VERSION = 12` inside the guest.


huapox:
can this be merged

PR #824, opened 2026-05-27
Fix DNS UDP response delivery to guests

enricoschaaf:
## Summary

- Inject forwarded DNS-over-UDP responses directly into the guest RX ring.
- Remove the DNS interceptor response channel that tried to write forwarded responses back through the smoltcp UDP socket.
- Reuse the UDP response frame construction path so DNS replies are delivered to the guest as actual Ethernet frames.

## Problem

The DNS interceptor reads guest UDP/53 packets from the smoltcp socket and forwards them asynchronously to the DNS forwarder. Before this change, the async UDP proxy sent successful upstream responses back over an internal response channel, and the interceptor then called `socket.send_slice(...)` on the smoltcp UDP socket.

That means DNS responses were re-entered through smoltcp as socket output after the original guest packet had already been consumed by the interception path. In practice, this can leave the guest resolver waiting even though the upstream query succeeded: the response is not injected into the guest-facing RX ring as a packet arriving from the gateway/resolver, so the guest can time out DNS lookups.

This is different from the generic UDP relay path, which already constructs a UDP response packet and enqueues the frame into the guest RX ring.

## Fix

This changes DNS-over-UDP forwarding to use the same guest-delivery model as the UDP relay:

- preserve the guest source endpoint from the intercepted DNS query,
- preserve the original destination address so the response appears to come from the resolver address the guest queried,
- construct a UDP response frame with the gateway and guest MAC addresses,
- enqueue that frame into `SharedState`'s RX ring and notify the libkrun network worker.

With this, forwarded DNS answers are delivered to the guest as incoming network frames instead of being written back through the interceptor's smoltcp socket.

## Testing

Not run, per request.

### DIFF
--- a/crates/network/lib/dns/interceptor.rs
+++ b/crates/network/lib/dns/interceptor.rs
@@ -1,10 +1,11 @@
 //! DNS query interception: the smoltcp ↔ channel bridge.
 //!
 //! `DnsInterceptor` owns the smoltcp UDP socket bound to `gateway:53`
-//! and a pair of channels to the async forwarder task spawned on the
-//! tokio runtime. Each poll-loop iteration, `process()` reads pending
-//! queries off the smoltcp socket and hands them to the forwarder,
-//! then writes any forwarded responses back to the socket.
+//! and a channel to the async forwarder task spawned on the tokio
+//! runtime. Each poll-loop iteration, `process()` reads pending queries
+//! off the smoltcp socket and hands them to the forwarder. Forwarded
+//! responses are injected directly into the guest RX ring by the UDP
+//! proxy.
 //!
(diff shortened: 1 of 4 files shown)


appcypher (approved):

enricoschaaf:
@appcypher it made sense for you? Talked about this to @toksdotdev I somehow was some to sometimes hit this issue and break dns forwarding. But this was mostly the best guess after iterating and analysing the blocked vm from both a host and client side with codex. It works flawlessly on my side but was not sure it was the correct fix.

appcypher:
Yes, it is useful, that is why I merged it in. There are some issues but I have decided to fix them in https://github.com/superradcompany/microsandbox/pull/836 instead of duplicating your work.

Issue #791, opened 2026-05-26
Large HTTPS POSTs fail from sandbox under TLS proxy, while the same requests succeed outside the sandbox


venbrinoDev:
## Summary

Inside microsandbox, small HTTPS POSTs succeed, but larger HTTPS POSTs can fail after upload or hang waiting for a response.

This is not specific to OpenRouter. I reproduced it with other HTTPS domains as well.

The same OpenRouter request body succeeds outside the sandbox on the host machine.

This points to a sandbox TLS/network/proxy issue rather than an application bug.

## Observed behavior

Inside sandbox:

- small HTTPS POSTs succeed
- large HTTPS POSTs can fail in different ways depending on destination:
  - EOF during read:
    - `OpenSSL SSL_read ... unexpected eof while reading`
  - timeout waiting for response:
    - `Operation timed out ... with 0 bytes received`

Outside sandbox:

- the same large OpenRouter request succeeds from the host machine
- both stream and non-stream requests return `200 OK`

### 1. OpenRouter

Target:
- `https://openrouter.ai/api/v1/chat/completions`

Request:
- OpenAI-compatible JSON body
- around `135-137 KB`

Inside sandbox:
- request uploads fully
- connection closes after ~16-18s
- client sees EOF / fetch failure

Outside sandbox:
- same payload succeeds with `200 OK`

### 2. Other HTTPS domains

From inside sandbox with `curl -k` and an `Authorization: Bearer ...` header:

- `https://postman-echo.com/post`
  - ~1 KB body: `200 OK`
  - ~135 KB body: EOF during read
- `https://httpbin.org/post`
  - ~1 KB body: `200 OK`
  - ~135 KB body: timeout after 25s with 0 bytes received

This suggests the issue is not specific to OpenRouter or the OpenRouter API key.

## Additional TLS/certificate issue

Normal TLS validation inside the sandbox also appears unhealthy for many domains.

These fail with:

- `SSL certificate problem: certificate is not yet valid`

Observed on:
- `postman-echo.com`
- `httpbin.org`
- `github.com`
- `google.com`
- `example.com`

But:
- `openrouter.ai` validates cleanly

So there may also be a broader certificate/time/proxy problem in the sandbox environment.

## Why this looks sandbox-specific

- same large OpenRouter request fails only inside sandbox
- same request succeeds on host
- large POST failures also reproduce on non-OpenRouter domains
- small POSTs succeed
- failure pattern is size-sensitive and transport-level

[...]

PR #833, opened 2026-05-28
fix(runtime): sync guest clock with host's

### LINKED ISSUE #831: Guest clock not corrected after host sleep on macOS (no wake-event passthrough)
## Summary

On macOS, the guest clock freezes during host sleep and is never corrected on wake. After a 2-hour laptop sleep, `date` inside
the guest is 2 hours behind real time, breaking anything time-sensitive — most visibly TLS validation, which fails with
`certificate is not yet valid` because msb's intercepted-CA certs are issued with `notBefore = now()` at TLS-handshake time.

## Repro

1. Boot a sandbox on macOS (any image with `curl`).
2. Note `date` in guest matches host.
3. Close laptop lid (or `pmset sleepnow`) for 30+ minutes.
4. Wake host.
5. In guest: `date` — still shows pre-sleep time.
6. `curl https://google.com` — fails with:
SSL certificate problem: certificate is not yet valid
issuer: CN=microsandbox CA; O=microsandbox
Because the intercept cert's `notBefore` is in the guest's "future" (real now), but the guest's clock is in the past.

## Root cause

macOS path uses libkrun + Hypervisor.framework. HVF has no `ptp_kvm` equivalent. The guest's TSC stops counting while the host
CPU is suspended, and on wake nothing tells the guest to re-stamp its clock. The virtualized RTC is set once at boot and then
ticks from the (frozen) TSC.

## Current workaround (figbox)

Running a stateless `chronyd -q` (one-shot step + exit) every 60s from inside the guest. Bounds max drift to ~60s, works for
any drift magnitude because each poll has no prior frequency model to fight. A long-lived `chronyd` daemon does *not* work —
its frequency estimator flags the huge post-wake offset as an outlier and refuses to step.

## Environment

- microsandbox 0.4.6 / sdk-go 0.4.7-rc1
- macOS (Apple Silicon), Hypervisor.framework backend

toksdotdev:
## tl;dr

sync guest wall clocks from the host so sandboxes recover from host sleep, suspend, and other pauses without breaking time-sensitive guest workloads.

fixes #831.

## description

- add a `core.clock.sync` protocol message carrying host unix time in nanoseconds
- send clock sync frames from the runtime relay on startup, every 60 seconds, and after a 6 second host wall-clock jump
- apply sync in agentd only when guest realtime differs by more than 100ms
- keep clock sync messages from refreshing the sandbox idle timer
- move host-side clock sync code into `crates/runtime/lib/clock.rs`


venbrinoDev:
@toksdotdev do you know if this fix addresses the issue reported in https://github.com/superradcompany/microsandbox/issues/791?

toksdotdev:
@venbrinoDev no, it doesn't. but i have a potential fix for #791 which i'll put up in a bit.

PR #834, opened 2026-05-28
fix(network): flush tls intercept upstream writes

### LINKED ISSUE #791: Large HTTPS POSTs fail from sandbox under TLS proxy, while the same requests succeed outside the sandbox

toksdotdev:
## tl;dr

flush the upstream tls stream after forwarding decrypted guest plaintext in the tls intercept proxy.

## description

- fixes #791
- tracks whether guest plaintext was forwarded to the upstream server
- flushes the `tokio-rustls` client stream after each drained plaintext batch so large request bodies are not left buffered
- keeps large tls-intercepted posts from timing out or ending with eof while the upstream server waits for the body


theycallmeswift:
@toksdotdev can we get a 0.5.3 release with this fix please?

appcypher:
> @toksdotdev can we get a 0.5.3 release with this fix please?

working on it

PR #858, opened 2026-05-31
fix(network/dns): return NXDOMAIN for policy-denied queries so stub resolvers fail closed

dijdzv:
## Problem

In a deny-by-default sandbox, a DNS lookup for a denied name **hangs** the
guest for tens of seconds. The forwarder answers policy-denied queries with
`RCODE=REFUSED`, but stub resolvers (e.g. glibc `getaddrinfo`) don't
fail-fast on `REFUSED` — the lookup blocks even though the egress is
correctly denied. (Connecting to a denied IP is unaffected; only DNS *name*
resolution hangs.)

Repro (egress allowing only one host):

```
msb exec <sb> -- getent hosts example.com      # hung
msb exec <sb> -- curl -m5 https://example.com  # -m5 never fires
```

## Fix

Answer policy denials with `NXDOMAIN` — an authoritative negative, so the
stub fails immediately (the convention DNS blockers like Pi-hole use). Covers
all three synthesized-denial paths in `DnsForwarder::forward` (block list,
denied resolver IP, rebind protection). `SERVFAIL` is left for genuine
upstream failures.

## Verification

After: `getent` fails in <1s (RC=2), `curl -m5` exits 6 immediately; allowed
domains still resolve. Added `build_status_response_nxdomain_variant`;
`cargo test -p microsandbox-network` passes.


dijdzv:
Honestly, I don't fully understand the changes needed to fix the tests, so I'm not confident about them.

toksdotdev:
hey @dijdzv , yeah this is a valid change. i thought i had updated these to `NXDOMAIN` as part of #578, but seems i never made the change.

lgtm! pushed a few more changes to the branch. thanks for the fix.

toksdotdev (approved):

PR #864, opened 2026-05-31
feat(sandbox): per-sandbox labels and label-based selection

### LINKED ISSUE #748: Sandbox tagging
Using the Go SDK, I'm developing an application that will effectively be a tenant of Microsandbox. My application must assume there will be other tenants, so it can't assume that it owns all sandboxes.

My application needs some way to identify the sandboxes it owns. One way to do this would be to track ownership externally such as in a database or JSON file, but this could become desynced and takes more work to implement. Another way could be a sandbox name prefix.

I'm proposing a feature for tagging sandboxes. The tags can be used as selectors both in the SDK and CLI.

Tags could be useful in the CLI to enable mass operations. During development I've found it inconvenient to name each individual sandbox to stop/start/remove. Tags could enable you to do something like:

```
msb stop --tags engine
msb start --tags engine
msb remove --tags engine
```

This way I can quickly perform operations only on sandboxes owned by my application.

In the SDK only one method would need to be added - to list sandboxes by tag. Stop/start/remove operations by tag could be implemented by the consumer.

```go
sandboxes, err := msb.ListSandboxesByTags(t.Context(), []string{"engine"})
```` 

The sandboxes returned would have all of the tags in the string slice.

Meta tags could also be useful. Meaning ones that are automatically assigned by Microsandbox. For example, there could be meta tags based on sandbox state which would enable this:

```
msb stop --tags engine,state:running
msb start --tags engine,state:stopped
msb remove --tags engine,state:stopped
msb remove --tags engine,state:crashed
```

### PR DESCRIPTION
**Pull request by toksdotdev**
adds per-sandbox labels (`key=value` metadata set at creation) and the ability to select sandboxes by them.

- set labels via the rust/go/python/node builders, `--label` on `msb create`/`run`, or `msb.yaml`. a bare `--label gpu` is a valueless marker.
- select by label: `Sandbox::list_with(...)` in the sdks, and `--label` on `msb stop`/`start`/`rm`/`ls`/`ps`.

closes #748.

### CODE REVIEW

File: crates/cli/lib/commands/stop.rs [hunk 1]
```diff
@@ -13,9 +15,13 @@ use crate::ui;
 #[derive(Debug, Args)]
 pub struct StopArgs {
     /// Sandbox(es) to stop.
-    #[arg(required = true)]
```
appcypher, line 16: is there a reason for making it un-required?

toksdotdev: good catch. it was so a label-only invocation (`msb stop --label app=engine`) could parse without also requiring a name. 

but that's better enforced at the clap layer, so i've switched to `required_unless_present = "label"`. it's still required unless `--label` is given, and omitting both now yields a proper clap usage error instead of a runtime one.

PR #907, opened 2026-06-04
refactor(sdk/go)!: remove CreateSandboxDetached

cedws:
The justification for removing this is that it's a simple wrapper around CreateSandbox that appends the WithDetached() functional option, increasing the API surface for little benefit.


appcypher:
This is a better design. I'm generalizing this to other SDKs where possible.

appcypher (approved):

PR #911, opened 2026-06-04
fix(agentd): restore default guest privileges

### LINKED ISSUE #908: Mount errors with docker in sandbox
I am following the [Docker in a sandbox](https://docs.microsandbox.dev/recipes/docker/docker-in-sandbox) recipe. Unfortunately, it doesn't work on my system:

```
$ uname -a
Linux tims-framework 6.17.0-35-generic #35~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 19:30:42 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

$ msb --version
msb 0.5.4

$ msb run --name docker-demo --replace \
        --memory 2G \
        --tmpfs /tmp:1G \
        --entrypoint sh \
        --script-path start-docker:./start-docker.sh \
        docker:dind
/ # start-docker 
Starting Docker. ready
/ # docker run --rm hello-world
Unable to find image 'hello-world:latest' locally
latest: Pulling from library/hello-world
4f55086f7dd0: Extracting 1 s
d5e71e642bf5: Download complete 
docker: failed to extract layer (application/vnd.oci.image.layer.v1.tar+gzip sha256:4f55086f7dd096d48b0e49be066971a8ed996521c2e190aa21b2435a847198b4) to overlayfs as "extract-110317261-24HR sha256:897b3f2a7c1bc2f3d02432f7892fe31c6272c521ad4d70257df624504a3238b4": failed to unmount /tmp/docker/containerd/daemon/tmpmounts/containerd-mount1703962267: operation not permitted: failed to mount /tmp/docker/containerd/daemon/tmpmounts/containerd-mount1703962267: mount source: "/tmp/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/1/fs", target: "/tmp/docker/containerd/daemon/tmpmounts/containerd-mount1703962267", fstype: bind, flags: 20480, data: "", err: operation not permitted

Run 'docker run --help' for more information
/ # mount
overlay on / type overlay (rw,relatime,lowerdir=/.msb/rootfs/lower,upperdir=/.msb/rootfs/upperfs/upper,workdir=/.msb/rootfs/upperfs/work,uuid=on)
msb_runtime on /.msb type virtiofs (rw,relatime)
devtmpfs on /dev type devtmpfs (rw,relatime,size=1036340k,nr_inodes=259085,mode=755)
proc on /proc type proc (rw,nosuid,nodev,noexec,relatime)
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
cgroup2 on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime)
devpts on /dev/pts type devpts (rw,nosuid,noexec,relatime,mode=600,ptmxmode=000)
tmpfs on /dev/shm type tmpfs (rw,nosuid,noexec,relatime)
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,relatime,size=1048576k)
```

appcypher:
## TL;DR
Restore normal guest-root behavior by default so sudo and Docker-in-Docker work again, while keeping the stronger hardening available through an explicit restricted security profile.

Closes #908 by restoring the default guest mount privileges Docker-in-Docker needs for containerd layer extraction and overlay/bind mounts.

## Description
- Stop applying no_new_privs and the CAP_SYS_ADMIN drop to every exec session by default.
- Add a `--security restricted` profile that keeps the previous hardening behavior for callers that want it.
- Thread `MSB_SECURITY_PROFILE` through runtime spawn, agentd boot config, and exec session handling.
- Make `nosuid` and `nodev` explicit mount flags for CLI volumes, tmpfs mounts, Rust mount builders, and the guest mount protocol.
- Update `msb inspect` so it shows `ro` or `rw` plus any explicit `noexec`, `nosuid`, or `nodev` flags.
- Update docs to explain normal default mount behavior, restricted profile tradeoffs, and Docker-in-Docker compatibility.

```bash
msb run --security restricted -v ./src:/app:ro,noexec,nosuid,nodev alpine
```

## Test Plan
- [x] `cargo fmt --all` exits 0
- [x] `git diff --check` exits 0
- [x] `cargo test -p microsandbox-cli --lib commands::common` passes
- [x] `cargo test -p microsandbox --lib runtime::spawn` passes
- [x] `cargo test -p microsandbox --lib sandbox::types` passes
- [x] `just build` builds patched agentd and signed `build/msb`


huksley:
Getting error when running docker, is it fixed in 0.5.5 or should I wait for another release  @appcypher 

```
docker: Error response from daemon: failed to mount /tmp/containerd-mount1489078651: mount source: "overlay", target: "/tmp/containerd-mount1489078651", fstype: overlay, flags: 0, data: "workdir=/var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/3/work,upperdir=/var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/3/fs,lowerdir=/var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/2/fs:/var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/1/fs,index=off", err: invalid argument
> dmesg
[   15.345446] overlay: filesystem on /var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/3/work not supported as upperdir
```

appcypher:
@huksley I don’t think this PR is what solves the overlay-on-overlay problem.

AFAIK, overlay-on-overlay mounts are not supported by linux. What you want instead is for Docker’s data root to live on a non-overlay mount, such as a mounted disk image.

##

The next release will bring support for disk-image-backed volumes: https://github.com/superradcompany/microsandbox/pull/927. So the recommended path is to create one and mount it as `/var/lib/docker`:

```bash
msb volume create docker-data --kind disk --size 20G
```

```bash
msb run --name docker-demo --replace \
  --memory 2G \
  --mount-named docker-data:/var/lib/docker \
  --script start='dockerd >/tmp/dockerd.log 2>&1 &
  timeout 60 sh -c "until docker info>/dev/null 2>&1;do sleep 1;done" || {
    cat /tmp/dockerd.log
    false
  }
  exec sh' \
  --entrypoint start \
  docker:dind
```

Then, inside the sandbox, you can run:

```bash
docker run -it --rm ubuntu bash
```

Personally, I’d recommend this approach. In the future, we’ll likely support creating named volumes on first mount, so you won’t need the separate `volume create` step. In fact, I will fix that right now before the next release.

##

The other, less efficient alternative is to use Docker’s `vfs` storage driver:

```bash
msb run --name docker-demo --replace \
  --memory 2G \
  --script start='dockerd --storage-driver=vfs >/tmp/dockerd.log 2>&1 &
  timeout 60 sh -c "until docker info>/dev/null 2>&1;do sleep 1;done" || {
    cat /tmp/dockerd.log
    false
  }
  exec sh' \
  --entrypoint start \
  docker:dind
```

appcypher:
@huksley in the latest release v0.5.6, you can now run docker within a sandbox in a single command

```sh
msb run --name docker-demo --replace \
  --memory 2G \
  --mount-named docker-data:/var/lib/docker:kind=disk,size=10G \
  --script start='dockerd >/tmp/dockerd.log 2>&1 &
  timeout 60 sh -c "until docker info>/dev/null 2>&1;do sleep 1;done" || {
    cat /tmp/dockerd.log
    false
  }
  exec sh' \
  --entrypoint start \
  docker:dind
```

PR #930, opened 2026-06-06
fix(network): wake publisher after draining host channel

chiehwangs:
## Description

Published-port responses can stall behind a slow host-side client because the inbound relay only wakes the poll loop after the host socket write completes.

## Root Cause

Receiving from `to_host_rx` frees capacity for more smoltcp-to-host data, but the wake happened after `write_all().await`. When that write blocks, the poll loop is not notified that it can refill the channel.

## Fix

Wake the poll loop immediately after draining `to_host_rx`, before awaiting the host socket write, and add a regression test that covers the blocked-write timing.

### DIFF
--- a/crates/network/lib/publisher.rs
+++ b/crates/network/lib/publisher.rs
@@ -761,12 +761,14 @@ async fn inbound_relay_task(
             data = to_host_rx.recv() => {
                 match data {
                     Some(bytes) => {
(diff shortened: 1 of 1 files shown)


toksdotdev:
hey, @chiehwangs. appreciate the fix. lgtm!

toksdotdev (approved):

PR #940, opened 2026-06-07
fix(network): substitute secrets over plain HTTP

### LINKED ISSUE #646: support secret substitution for plain http
secret substitution only runs in the tls interception path, so plain http requests forward the placeholder verbatim. `Secret.env(...)` does not work for plain http endpoints.

TheRealSibasishBehera:
Closes #646.

SecretsHandler was only wired into the TLS path. Plain HTTP connections went through a raw relay — placeholders reached the server unchanged, and violation detection (blocking sends to disallowed hosts) never ran.

What changed

  - proxy.rs - after confirming a connection is not TLS, accumulate chunks until `\r\n\r\n,` extract Host: via `httparse`, build a `SecretsHandler`, and substitute through both the first-flight replay and the relay loop.
  - handler.rs - new `SecretsHandler::new_plain_http(): tls_intercepted = false` so the `require_tls_identity` gate stays active, DNS-cache pin and per-request Host authority check both enforced.

Opt-in required  plain HTTP substitution requires . `require_tls_identity(false)` per secret. The default is true, so no existing secret changes behaviour. Violation detection runs for all secrets regardless.


TheRealSibasishBehera:
@appcypher I did a rebase and the tests should be passing now. Could you take a look?

toksdotdev:
hey @TheRealSibasishBehera. appreciate the pr. 🙌🏽  i will give it a review later today.

crates/network/lib/proxy.rs
```diff
@@ -125,16 +132,49 @@ async fn tcp_proxy_task(
...
-        return Ok(());
+    // Replay the buffered first flight — run through secrets handler first.
+    if !initial_buf.is_empty() {
+        let out = match secrets_handler.as_mut() {
+            Some(h) => match h.substitute(&initial_buf) {
+                Ok(cow) => cow.into_owned(),
+                Err(action) => {
+                    tracing::warn!(dst = %connect_dst, violation = ?action, "secret violation in first flight");
+                    return Ok(());
+                }
```
toksdotdev, line 169: we should handle the `BlockAndTerminate` violation action like we do [here](https://github.com/superradcompany/microsandbox/blob/main/crates/network/lib/tls/proxy.rs#L398-L405).

crates/network/lib/proxy.rs
```diff
@@ -145,21 +185,37 @@ async fn tcp_proxy_task(
...
-                        if let Err(e) = server_tx.write_all(&bytes).await {
-                            tracing::debug!(dst = %connect_dst, error = %e, "write to server failed");
-                            break;
+                        let out = match secrets_handler.as_mut() {
+                            Some(h) => match h.substitute(&bytes) {
+                                Ok(cow) => cow.into_owned(),
+                                Err(action) => {
+                                    tracing::warn!(dst = %connect_dst, violation = ?action, "secret violation");
+                                    break;
+                                }
```
toksdotdev, line 206: same here; we should handle the block and terminate violation action.

crates/network/lib/proxy.rs
```diff
@@ -125,16 +132,49 @@ async fn tcp_proxy_task(
...
+    // non-TLS byte and may return before headers are complete. Keep reading until
+    // \r\n\r\n so extract_http_host always sees a full header block.
+    if !is_tls && !secrets.secrets.is_empty() {
+        initial_buf =
+            peek_for_http_headers(initial_buf, &mut from_smoltcp, PEEK_BUF_SIZE, PEEK_BUDGET).await;
+    }
+
+    let mut secrets_handler: Option<SecretsHandler> = if !secrets.secrets.is_empty() && !is_tls {
+        extract_http_host(&initial_buf)
+            .map(|host| SecretsHandler::new_plain_http(&secrets, &host, guest_dst.ip(), &shared))
```
toksdotdev: this would mean requests with no host headers, malformed headers, or un-parseable header would bypass secret substitution. we should still subject such requests regardless to secret substitution. so something like this:

```rust
  let mut secrets_handler = if !secrets.secrets.is_empty() && !is_tls {
      Some(match extract_http_host(&initial_buf) {
          Some(host) => SecretsHandler::new_plain_http(&secrets, &host, guest_dst.ip(), &shared),
          None => SecretsHandler::new_plain_http_invalid_host(&secrets),
      })
  } else {
      None
  };
```

the `::new_plain_http_invalid_host()` i proposed would make all secrets as ineligible for substitution since the host can't be proven. maybe only grant an exception if all the secrets have a `HostPattern::Any`, in which case it's safe to allow through.

crates/network/lib/proxy.rs
```diff
@@ -125,16 +132,49 @@ async fn tcp_proxy_task(
...
 
+    let is_tls = initial_buf.first() == Some(&0x16);
+
+    // For plain-HTTP connections with secrets, peek_for_sni bails on the first
+    // non-TLS byte and may return before headers are complete. Keep reading until
+    // \r\n\r\n so extract_http_host always sees a full header block.
+    if !is_tls && !secrets.secrets.is_empty() {
+        initial_buf =
+            peek_for_http_headers(initial_buf, &mut from_smoltcp, PEEK_BUF_SIZE, PEEK_BUDGET).await;
+    }
```
toksdotdev: this peeks for http headers whenever any secret exists (`!secrets.secrets.is_empty()`), but most secrets are tls-only by default (`require_tls_identity = true`) and can’t be substituted on plain http. 

that means unrelated non-tls protocols in a sandbox with default secrets can now stall until `\r\n\r\n`, the cap, channel close, or the 5s `PEEK_BUDGET`.

can we gate the http-header peek on whether plain http substitution is actually possible maybe a `SecretsConfig::has_plain_http_candidates()` helper that checks for at least one secret with `require_tls_identity == false` and an enabled injection scope. so something like:

```rust
pub(crate) fn has_plain_http_candidates(&self) -> bool {
    self.secrets.iter().any(|secret| {
        !secret.require_tls_identity
            && (secret.injection.headers
                || secret.injection.basic_auth
                || secret.injection.query_params
                || secret.injection.body)
       })
}
```

domain-policy sni peeking should stay as-is, but for non-tls traffic with no plain-http candidates we should replay the first flight immediately instead of waiting for http headers.

toksdotdev (changes requested):
my honest apologies for the delay on this. got swamped yesterday with a few other things.

crates/network/lib/proxy.rs
TheRealSibasishBehera: Done in 5d9d103a

crates/network/lib/proxy.rs
TheRealSibasishBehera: Done in 5d9d103a

crates/network/lib/proxy.rs
TheRealSibasishBehera: Done in 332739d0

crates/network/lib/proxy.rs
TheRealSibasishBehera: Why all-or-nothing for the `HostPattern::Any` ? would per-secret check for `HostPattern::Any` (substitute the Any secrets, block the rest) make more sense ?

toksdotdev: so we do per-secret policy checks today when the host can be inferred.

however, for the no-host/invalid-host path, i mainly us want to avoid making host-scoped secrets eligible **when we can’t prove the host**. the only way to kinda guarantee that the request should be allowed through is if all the secret's policy has `HostPattern::Any`.

TheRealSibasishBehera: Makes sense
 Done in 29dd7f28, with e2e coverage in b2f3685a.

TheRealSibasishBehera:
thanks for the review @toksdotdev! 
I have addressed all the review comments and merged the latest main in. ready for another look.

toksdotdev (approved):
awesome! thanks for making the changes. 🙏🏽

PR #977, opened 2026-06-13
fix(agentd): disable DAD for IPv6 addresses in virtual network

### LINKED ISSUE #948: IPv6 connectivity broken - guest address stays tentative indefinitely
Sandbox guests are provisioned with an IPv6 address (fd42:6d73:62:N::2/64) but outbound IPv6 traffic never works. The address stays permanently tentative and the kernel drops all outbound IPv6 packets.

Digging into this, the issue appears to be in how the virtual network gateway handles NDP — specifically the Router Solicitation the guest sends at boot. Without a Router Advertisement in response, the guest's IPv6 stack never fully initializes and DAD (Duplicate Address Detection) never completes.

IPv4 does not have this problem — the address is marked preferred immediately on assignment.

Reproducer

Branch [test/ipv6-connectivity-e2e](https://github.com/TheRealSibasishBehera/microsandbox/commit/7eeb2fede3078a36849ccf3333dc5cbd6547b918) has end-to-end tests that demonstrate the failure:

  `cargo nextest run -p microsandbox --tests --run-ignored=only -E 'test(ipv6)'`

Tests cover TCP connect, HTTP over IPv6, ICMPv6 ping, and gateway reachability.

TheRealSibasishBehera:
Closes #948.

## Problem

Sandbox guests on IPv6-capable hosts are provisioned with a ULA address
(`fd42:6d73:62:N::2/64`) but the address stays permanently `tentative`
and the kernel drops all outbound IPv6 traffic.

The root cause is that smoltcp has no `RouterSolicit` handler — without a
Router Advertisement the guest's DAD state machine never completes.

## Approach

Set `IFA_F_NODAD` on the `RTM_NEWADDR` netlink message for IPv6 in agentd.
The kernel skips DAD and immediately marks the address `PREFERRED`.

Kata Containers applies [the same flag for the same reason](https://github.com/kata-containers/agent/blob/master/network.go#L136-L142)
- in a single-tenant virtual network the orchestration layer owns address
assignment, so uniqueness is guaranteed by construction. Each sandbox slot
owns a deterministically assigned prefix (`fd42:6d73:62:N::/64`); no peer
can hold a duplicate.

## Alternate approach considered

Implementing a `RouterSolicit` → `RouterAdvertisement` handler at the
gateway . Protocol-correct but requires synthesising raw
ICMPv6 frames outside smoltcp's abstraction with no practical benefit given
the uniqueness guarantee the slot allocator already provides.

### DIFF
--- a/crates/agentd/lib/network.rs
+++ b/crates/agentd/lib/network.rs
@@ -409,7 +409,7 @@ mod linux {
         unsafe {
             (*ifa).ifa_family = family;
             (*ifa).ifa_prefixlen = prefix_len;
-            (*ifa).ifa_flags = 0;
+            (*ifa).ifa_flags = if is_ipv4 { 0 } else { libc::IFA_F_NODAD as u8 };
             (*ifa).ifa_index = ifindex;
             (*ifa).ifa_scope = libc::RT_SCOPE_UNIVERSE;
         }
--- a/crates/microsandbox/tests/host_access.rs
+++ b/crates/microsandbox/tests/host_access.rs
@@ -358,3 +358,37 @@ async fn group_host_deny_blocks_host_only() {
 
     teardown(sb, name).await;
 }
+
+/// Guest ULA IPv6 address is present and in preferred state on boot.
+///
+/// Skipped when the host has no IPv6 route — the runtime does not provision
+/// a guest IPv6 address on IPv4-only hosts.
+#[msb_test]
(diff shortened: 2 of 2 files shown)


toksdotdev (approved):
really appreciate the fix. lgtm!

`IFA_F_NODAD` in this case makes sense. these addresses are manually and uniquely assigned, so there's no chance for collision to warrant a DAD.

PR #1020, opened 2026-06-23
feat(filesystem): add guest-write quotas for virtiofs mounts

### LINKED ISSUE #1007: /.msb Runtime Directory Has No Storage Quota and Shows Inconsistent Space Usage
# Bug Report: /.msb Runtime Directory Has No Storage Quota and Shows Inconsistent Space Usage

## Description

The `/.msb` runtime directory inside the sandbox VM has no storage quota limits, allowing unlimited writes that consume host disk space. Additionally, there is a significant discrepancy between `du` and `df` space usage reports.

## Environment

- **microsandbox version**: v0.5.7 (based on changelog)
- **OS**: Debian GNU/Linux 13 (trixie) - sandbox guest
- **Host**: Windows 10
- **Kernel**: Linux sb-alice 6.12.68

## Steps to Reproduce

1. Create a sandbox instance
2. Write large amounts of data to `/.msb/` directory
3. Observe that no quota limit is enforced
4. Compare `du` and `df` output

```bash
# Write 5GB to /.msb
dd if=/dev/urandom of=/.msb/test.bin bs=1M count=5120

# Check space usage
du -sh /.msb/      # Shows actual file size
df -h /.msb        # Shows different (higher) usage
```

## Expected Behavior

1. `/.msb` should have configurable storage quota (like `upper_size_mib` for overlay)
2. `du` and `df` should show consistent space usage
3. Writing beyond quota should return "No space left on device" error

## Actual Behavior

1. **No quota limit**: Successfully wrote 60GB+ data to `/.msb/` without any restriction
2. **Inconsistent space reporting**:
   - `du -sh /.msb/` shows actual file size (e.g., 15GB)
   - `df -h /.msb` shows much higher usage (e.g., 52GB)
   - After deleting files, `df` shows space reclaimed but discrepancy remains
3. **Host disk consumption**: Data written to `/.msb/` actually consumes host disk space (verified by monitoring host storage)

### Storage Quota Test

| Location | Limit | Can Write Beyond? |
|----------|-------|-------------------|
| `/workspace` | 944MB | ❌ No |
| `/tmp` | 128MB | ❌ No |
| `/` (overlay) | 3.9GB | ❌ No |
| `/.msb` | None | ✅ Yes (60GB+) |

### Space Reporting Discrepancy

```
Before writing 5GB:
  du: 0 bytes
  df: 52GB used

After writing 5GB:
  du: 5GB
  df: 57GB used (increased by 5GB)

After deleting file:
  du: 0 bytes
  df: 52GB used (decreased by 5GB, but 52GB still unexplained)
```

## Why This Matters

1. **Resource isolation**: Users can bypass storage limits by writing to `/.msb/`
2. **Host stability**: Unbounded writes can fill host disk
3. **Monitoring confusion**: Inconsistent `du`/`df` makes capacity planning difficult

## Suggested Fix

1. Add `msb_size_mib` or similar config option in `sandbox_defaults` to limit `/.msb/` storage
2. Consider mounting `/.msb/` with quota support (e.g., virtiofs quota, or separate block device)
3. Fix the space reporting discrepancy between `du` and `df`

## Additional Context

The `/.msb/` directory is essential for sandbox operation (contains rootfs overlay layers), so it cannot be disabled. However, it should have configurable resource limits like other partitions.

```json
// Suggested config addition
{
  "sandbox_defaults": {
    "oci": {
      "upper_size_mib": 4096
    },
    "msb_size_mib": 1024  // NEW: limit /.msb storage
  }
}
```

appcypher:
## TL;DR
PassthroughFs now caps how much a guest can write to a virtiofs mount, so a sandbox can no longer fill the host disk through `/.msb` or a bind mount. Fixes #1007.

## Description
- `/.msb` gets a fixed 16 MiB guest-write cap. It stays a virtiofs control channel (the host reads `heartbeat.json` and writes scripts/TLS certs through it) rather than becoming a block device, so the cap is not user-configurable.
- Bind mounts get a default 4 GiB guest-write quota, overridable per mount.
- Named directory volumes now enforce their existing `quota_mib`, which was stored but ignored before.
- Enforcement charges only file growth (new EOF minus current size), not write volume, so a file rewritten in place like the 1 KiB/s heartbeat never drains the budget.
- Pre-existing files are never charged. That baseline is walked lazily on the guest's first write, so a heavy bind mount adds nothing to boot and a read-only workload never walks the tree.
- Guest `df` on a quota'd mount now reports the budget (total = baseline + quota, used tracks `du`) instead of the host's full disk.

Rust SDK:
```rust
let sandbox = Sandbox::builder()
    .image("python:3.12")
    .volume("/data", |m| m.bind("./data").quota(2.gib()))
    .build()
    .await?;
```

CLI (default applies when omitted, `quota=` is in MiB):
```bash
msb run python --mount-dir ./cache:/cache:quota=512
msb run python -v ./data:/data:quota=2048
```

Python (mounts are keyed by guest path):
```python
from microsandbox import MountConfig, MountKind

volumes = {"/data": MountConfig(kind=MountKind.BIND, bind="./data", quota_mib=2048)}
```

Node:
```ts
sandbox.volume("/data", (m) => m.bind("./data").quota(2048));
```

## Test Plan
- [x] `cargo test -p microsandbox-filesystem` passes (quota unit + end-to-end FUSE invariants, 607 tests)
- [x] `cargo test -p microsandbox-runtime -p microsandbox-types -p microsandbox -p microsandbox-cli` passes (605 tests)
- [x] `cargo clippy --workspace -- -D warnings` is clean
- [x] `cargo doc --workspace --no-deps` is clean
- [x] live check: `dd` past 16 MiB into `/.msb` returns ENOSPC and `df /.msb` shows the cap (verified in a booted alpine sandbox)

### DIFF
--- a/crates/cli/lib/commands/common.rs
+++ b/crates/cli/lib/commands/common.rs
@@ -755,6 +755,7 @@ pub fn apply_volume(builder: SandboxBuilder, spec: &str) -> anyhow::Result<Sandb
         spec,
         CliMountOptionSupport {
             policies: true,
+            quota: true,
             ..CliMountOptionSupport::default()
         },
     )?;
@@ -764,10 +765,23 @@ pub fn apply_volume(builder: SandboxBuilder, spec: &str) -> anyhow::Result<Sandb
     let guest = parsed.guest.to_string();
     let options = parsed.options;
     Ok(builder.volume(guest, move |mut m| {
+        let quota_mib = options.quota_mib;
         m = if is_path {
-            m.bind(&source)
+            let mut b = m.bind(&source);
+            if let Some(q) = quota_mib {
+                b = b.quota(q);
+            }
+            b
         } else {
-            m.named_with(&source, |v| v.ensure_exists())
+            // A named volume routes its quota through the named sub-builder
+            // rather than the bind-only `.quota()`.
+            m.named_with(&source, move |mut v| {
+                v = v.ensure_exists();
+                if let Some(q) = quota_mib {
+                    v = v.quota(q);
+                }
+                v
+            })
         };
         if options.readonly {
             m = m.readonly();
@@ -829,6 +843,7 @@ pub fn apply_explicit_dir_mount(
         spec,
         CliMountOptionSupport {
             policies: true,
+            quota: true,
             ..CliMountOptionSupport::default()
         },
     )?;
@@ -976,6 +991,9 @@ fn apply_common_mount_options(mut mount: MountBuilder, options: CliMountOptions)
     }
     if let Some(hp) = options.host_permissions {
         mount = mount.host_permissions(hp);
+    }
+    if let Some(quota) = options.quota_mib {
+        mount = mount.quota(quota);
     }
     mount
 }
--- a/crates/cli/lib/commands/inspect.rs
+++ b/crates/cli/lib/commands/inspect.rs
@@ -153,10 +153,17 @@ pub async fn run(args: InspectArgs) -> anyhow::Result<()> {
                         options,
                         stat_virtualization,
                         host_permissions,
+                        quota_mib,
                     } => {
                         let flags = mount_flags_suffix(*options);
                         let suffix = mount_policy_suffix(*stat_virtualization, *host_permissions);
-                        println!("  {guest:<16}\u{2192} {}{flags}{suffix}", host.display());
+                        let quota = quota_mib
+                            .map(|mib| format!(" [quota={mib}MiB]"))
+                            .unwrap_or_default();
+                        println!(
+                            "  {guest:<16}\u{2192} {}{flags}{suffix}{quota}",
+                            host.display()
+                        );
                     }
                     VolumeMount::Named {
                         name,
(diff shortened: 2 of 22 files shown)


appcypher:
Verified live in a booted alpine sandbox (msb 0.5.8, debug build):

```
### df /.msb BEFORE:
Filesystem                Size      Used Available Use% Mounted on
msb_runtime              16.0M         0     16.0M   0% /.msb
### writing 25 MiB to /.msb (cap is 16 MiB):
dd: error writing '/.msb/test.bin': No space left on device
17+0 records in
16+0 records out
16777216 bytes (16.0MB) copied, 0.037139 seconds, 430.8MB/s
### dd exit code: 1
### resulting file size:
-rw-r--r--    1 root     root      16777216 /.msb/test.bin
16.0M	/.msb/test.bin
### df /.msb AFTER:
Filesystem                Size      Used Available Use% Mounted on
msb_runtime              16.0M     16.0M         0 100% /.msb
```

Matches #1007's asks: quota enforced (ENOSPC at exactly 16 MiB), and du/df now agree (both 16.0M).

PR #1022, opened 2026-06-23
fix(network): substitute secrets through CONNECT proxies

### LINKED ISSUE #752: Secret substitution fails with HTTPS_PROXY that uses http:// and port 80
Running microsandbox in an environment with a corporate proxy configured as 

`HTTPS_PROXY="http://proxy.corporate.com:80"`

secret substitution for https connections fails as the TLS proxy isn't used.

For example,

`curl -v -H "Authorization: Basic $BASIC_AUTH" https://internal.corporate.com/api/auth`

gives output like

```log
* Uses proxy env variable NO_PROXY == '127.0.0.1,localhost'
* Uses proxy env variable HTTPS_PROXY == 'http://proxy.corporate.com:80'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying x.x.x.x:80...
* Connected to proxy.corporate.com (x.x.x.x) port 80 (#0)
* allocate connect buffer!
* Establish HTTP proxy tunnel to internal.corporate.com:443
> CONNECT internal.corporate.com:443 HTTP/1.1
> Host: internal.corporate.com:443
> User-Agent: curl/7.76.1
> Proxy-Connection: Keep-Alive
>
< HTTP/1.1 200 Connection Established
<
* Proxy replied 200 to CONNECT request
* CONNECT phase completed!

<snipped SSL cert verification>

> GET /api/auth HTTP/1.1
> Host: internal.corporate.com
> User-Agent: curl/7.76.1
> Accept: */*
> Authorization: Basic $MSB_ARTIFACTORY_BASIC_AUTH
>
* TLSv1.2 (IN), TLS header, Unknown (23):
{ [5 bytes data]
* Mark bundle as not supporting multiuse
< HTTP/1.1 401 Unauthorized
< Date: Mon, 18 May 2026 08:52:34 GMT
< Content-Type: application/json;charset=ISO-8859-1
< Content-Length: 108
< Connection: keep-alive

* Connection #0 to host proxy.corporate.com left intact
{
  "errors" : [ {
    "status" : 401,
    "message" : "Failed to decode basic authentication token"
  } ]
```

where `$MSB_ARTIFACTORY_BASIC_AUTH` is passed.

Perhaps related to #646

TheRealSibasishBehera:
Fixes https://github.com/superradcompany/microsandbox/issues/752#issuecomment-4767847704

## Summary

Secret substitution silently fails when the guest uses an HTTP CONNECT proxy (`HTTPS_PROXY=http://proxy:port`). The `SecretsHandler` DNS-cache pin check verifies that the guest previously resolved the SNI to the destination IP — but for CONNECT-tunnelled connections, the destination IP is the **proxy's**, not the target's. The guest never resolves `target.example.com → proxy-ip`, so the check always fails and secrets are never substituted.

## Changes

**`crates/network/lib/tls/proxy.rs`**
- Add `via_connect: bool` to `TlsProxyContext`; defaults to `false` in `spawn_tls_proxy`

**`crates/network/lib/secrets/handler.rs`**
- Add `new_tls_intercepted_via_connect`: skips the DNS-cache pin check. Safe because the proxy already cross-checked the ClientHello SNI against the CONNECT authority before this point.

**`crates/network/lib/proxy.rs`**
- Set `via_connect: expected_sni.is_some()` in `handle_connect_tunnel`
- For hostname CONNECT targets (`CONNECT api.example.com:443`), `via_connect: true` skips the now-redundant DNS-cache pin
- For IP-literal CONNECT targets (`CONNECT 1.2.3.4:443`), `expected_sni` is `None` so `via_connect: false` — the standard DNS-cache pin runs against the literal target IP, keeping a corroboration requirement in place

**`sdk/rust/tests/http_connect_secret.rs`**
- Two e2e regression tests using a local CONNECT proxy and `curl --proxytunnel`:
  1. Target host is never DNS-resolved by the guest before the proxied request
  2. Target host is DNS-resolved first; cache entry maps to gateway IP, not proxy IP

## Behaviour

| CONNECT target | SNI/authority check | DNS-cache pin |
|---|---|---|
| `CONNECT hostname:443` | runs, must match ClientHello SNI | skipped (authority already verified) |
| `CONNECT 1.2.3.4:443` | skipped (no hostname) | runs against `1.2.3.4` |
| Direct TLS (no CONNECT) | n/a | always runs |](https://github.com/superradcompany/microsandbox/issues/752#issuecomment-4767847704)


toksdotdev:
noticed the fix skipped the DNS pin by disabling `identity`, but that also accidentally disabled the check that ensures HTTP request host matches the TLS SNI. CONNECT should skip only the DNS pin. 

i pushed a small follow-up that keeps that host validation in place, so secrets are not substituted for mismatched request hosts.

asides that, lgtm! thank you.

toksdotdev (approved):

TheRealSibasishBehera:
thanks for the review and the follow‑up !

PR #1028, opened 2026-06-24
fix(runtime): mount named kind=disk volume as virtio-blk

zhuanyongxigua:
## TL;DR

Named volumes created with `kind=disk` were mounted over virtiofs instead of as a virtio-blk block device, so overlayfs (e.g. docker's default storage driver) on top of them failed with `EINVAL`. This routes them to virtio-blk, and also recovers the disk kind from the volume store so the fix survives a sandbox restart.

## Description

**1. Route named `kind=disk` volumes to virtio-blk** (`spawn.rs`, `VolumeMount::Named` arm)
- `sandbox_cli_args` discarded the volume's `create` kind and unconditionally called `push_dir_mount_arg` / `push_dir_mounts_spec`, mounting every named volume over virtiofs.
- A named `kind=disk` volume is provisioned on the host as an ext4 raw image (`disk.raw`), but mounting it over virtiofs means the guest never gets the block device. The kernel rejects overlayfs over virtiofs with `EINVAL` (`invalid argument` for docker-in-VM workloads).
- Now routes named `kind=disk` volumes through `push_disk_mount_arg` / `push_disk_mounts_spec` (virtio-blk, raw format, ext4 fstype), mirroring the `DiskImage` arm. Plain named directory volumes keep their virtiofs path and quota handling unchanged.

**2. Recover the disk kind from the volume store on restart**
- The `create` field is intentionally dropped when a sandbox config is persisted (the custom `Serialize` for `VolumeMount::Named` omits it). After stop/start from the DB, `create` is `None`, so the volume would regress to the virtiofs path — disk-backed workloads that work on first boot break on restart.
- Before rendering launch args, `spawn_sandbox` queries the volume store (`volume_entity`, `disk_format IS NOT NULL`) for named volumes whose kind isn't already known from `create`, and passes the resolved set into `sandbox_cli_args`. A volume is treated as disk when either `create` says so **or** the store reports it — so restarts mount the same virtio-blk ext4 image as the first boot.

## Test Plan

- `cargo test -p microsandbox --lib runtime::spawn::tests` — 46 passed, 0 failed. New tests:
  - `test_sandbox_cli_args_named_disk_volume`: named `kind=disk` renders `--disk` + `MSB_DISK_MOUNTS=...:fstype=ext4`, does not leak into `MSB_DIR_MOUNTS`.
  - `test_sandbox_cli_args_named_directory_volume`: plain named volume still mounts over virtiofs, no `--disk` / `MSB_DISK_MOUNTS`.
  - `test_sandbox_cli_args_named_disk_volume_restart_path`: `create == None` but the volume is disk-backed in the store → still virtio-blk.
- `cargo fmt -p microsandbox --check` — clean. `cargo clippy -p microsandbox --lib` — no warnings.
- Manual end-to-end on a real host: a patched `msb` ran a VM with `--mount-named <vol>:/var/lib/docker:kind=disk,size=8G`; in the guest `/var/lib/docker` is `ext4` on `/dev/vda` (not virtiofs), dockerd uses the `overlayfs` storage driver, and `docker run` succeeds (previously `invalid argument`).

## Follow-ups (out of scope here)

- The named-disk path passes a hardcoded `ext4` fstype to `push_disk_mounts_spec` (named disk volumes are always provisioned ext4 by `ensure_named_volumes`). If microsandbox ever supports other fstypes for named disk volumes, the restart path should carry the per-volume fstype (e.g. `db_disk_volumes` as a name→fstype map), matching how the `DiskImage` arm reads `fstype.as_deref()`.
- A mixed-volume test (disk + directory named volumes in one config) and a focused unit test for the store lookup would broaden coverage; the current tests cover the three core states (first-run disk / restart disk / directory).

### DIFF
--- a/sdk/rust/lib/runtime/spawn.rs
+++ b/sdk/rust/lib/runtime/spawn.rs
@@ -1309,22 +1309,36 @@ fn sandbox_cli_args(
                 host_permissions,
                 create,
             } => {
-                let vol_path = local.volume_path(name);
-                // Directory named volumes honor their configured quota. The
-                // value is available here when the volume is created or ensured
-                // this spawn (`create`); a quota set on a prior run and reloaded
-                // from a persisted config is not yet re-resolved from the store.
-                let quota_mib = create.as_ref().and_then(|c| c.quota_mib);
-                push_dir_mount_arg(
-                    &mut launch.mounts,
-                    guest,
-                    &vol_path.display(),
-                    *options,
-                    *stat_virtualization,
-                    *host_permissions,
-                    quota_mib,
-                );
-                push_dir_mounts_spec(&mut dir_mounts_val, guest, *options);
+                let is_disk = matches!(create, Some(c) if c.kind == VolumeKind::Disk);
+                if is_disk {
+                    let raw = local.volume_path(name).join("disk.raw");
+                    let id = guest_mount_tag(guest);
+                    push_disk_mount_arg(
+                        &mut launch.disks,
+                        &id,
+                        &raw.display(),
+                        &DiskImageFormat::Raw,
+                        *options,
+                    );
+                    push_disk_mounts_spec(&mut disk_mounts_val, &id, guest, Some("ext4"), *options);
+                } else {
+                    let vol_path = local.volume_path(name);
+                    // Directory named volumes honor their configured quota. The
+                    // value is available here when the volume is created or ensured
+                    // this spawn (`create`); a quota set on a prior run and reloaded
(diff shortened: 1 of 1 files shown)


appcypher:
thanks for the contribution. there's been a bunch of regressions in the past few days. 
some came from a massive [pr merge conflict resolution](https://github.com/superradcompany/microsandbox/pull/754) i made some days back. i'm finding and fixing all of them.

i will add my changes to your branch and merge.

PR #1126, opened 2026-07-06
docs: add a Reconfigure page and split the secrets page headings

toksdotdev:
two related guide changes for the modify/resize feature area.

**Reconfigure page (draft to evaluate).** a dedicated sandboxes page that gives the modify/resize model one home in guide voice: resize cpu/memory, update env/labels/workdir, rotate secrets, preview, and apply changes that need a restart. placed after lifecycle in the nav. if kept, the follow-up is to collapse the cli `msb modify` intro, the sdk `modify()` descriptions, and the lifecycle section into pointers here; dropping it is just deleting the page and its nav line.

**Secrets page headings.** added an 'add secrets when creating a sandbox' heading so the page's two distinct tasks, defining secrets at create time and changing them on a running sandbox, each show up in the on-this-page nav. previously only the running-sandbox heading existed.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -39,6 +39,7 @@
             "pages": [
               "sandboxes/overview",
               "sandboxes/lifecycle",
+              "sandboxes/reconfigure",
               "sandboxes/commands",
               "sandboxes/ssh",
               "sandboxes/filesystem",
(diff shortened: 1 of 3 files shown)


toksdotdev:
folded into #1124 per request to keep everything in one PR

PR #1127, opened 2026-07-06
docs: give the secrets page create-time and running-sandbox headings

toksdotdev:
the secrets page had a single heading (changing secrets on a running sandbox), so the create-time setup that fills most of the page had no entry in the on-this-page nav. added an 'add secrets when creating a sandbox' heading before the create example, leaving the placeholder-model intro as the page lead.


toksdotdev:
folded into #1126 per request to keep everything in one PR

PR #1150, opened 2026-07-08
perf(snapshot): store files with holes as GNU sparse tar entries on export

nuri-yoo:
## TL;DR

`snapshot export` streams the entire logical size of the sparse `upper.ext4` through tar and zstd, even when almost none of it is allocated: a default 4 GiB upper with ~2.9 MB of data takes ~6.5 s to export into a ~2 MB `.tar.zst`. This PR teaches export to write files that contain holes as old-GNU sparse tar entries, reading only the allocated extents. The same export drops to ~0.02 s and ~7 KB, and import reproduces both the logical content and the holes. Dense files keep byte-identical regular entries.

## Root cause

The upper layer is a sparse ext4 block image whose apparent size is the configured upper cap, not the data the sandbox wrote. `snapshot create` is already sparse-aware: it copies through `fast_copy`/`sparse_copy` (`crates/utils/lib/copy.rs`) walking `SEEK_DATA`/`SEEK_HOLE`, and is O(data). Export is not: `write_archive_entries` (`sdk/rust/lib/snapshot/archive.rs`) appends the upper with `append_path_with_name`, and the tar builder reads the file to EOF with a plain `io::copy`, so every hole is materialized as zeros and pushed through zstd. That makes export O(logical upper size). The module doc's "sparse files collapse cleanly under zstd" holds for the archive size, but not for the time spent producing it.

The create/export asymmetry is easy to see on any fresh sandbox — create runs in milliseconds, export in seconds, against the same upper:

```
msb create alpine --name prof
msb stop prof
time msb snapshot create snap --from prof
time msb snapshot export snap out.tar.zst
```

## Changes

- Export routes every file through a new `append_artifact_file`. On Unix, a `SEEK_DATA`/`SEEK_HOLE` walk (same idiom as `snapshot/verify.rs`) builds the allocation map, and files with holes are written as old-GNU sparse entries: GNU header with type `S`, `size` = archived bytes, `realsize` = logical size, and the inline 4-slot sparse map; chained 21-slot extended sparse headers for longer maps; then the data extents back to back, zero-padded to the tar block. Data runs are rounded to 512-byte blocks and merged, and a zero-length terminator marks a trailing hole — the same conventions GNU tar uses for this format.
- Files with no holes, empty files, non-regular files, filesystems without `SEEK_DATA` support, names that need a GNU long-name entry, and Windows all keep the existing dense `append_path_with_name` path and produce byte-identical archives to today.
- Import validation accepts `GNUSparse` alongside `Regular`/`Continuous`. No unpack changes are needed: tokio-tar already parses old-GNU sparse maps on read and unpacks holes with seek + `set_len`, so imported uppers come back sparse rather than zero-filled.
- `ExportOpts` and the SDK surfaces are unchanged.

## Compatibility

- The archive stays a plain tar (optionally zstd): old-GNU sparse is the sparse representation GNU tar and libarchive read. Verified that macOS bsdtar extracts a sparse export with identical content and reconstructs the holes.
- Importers without this change reject `GNUSparse` entries during validation ("archive contains unsupported entry type"), so an archive whose upper actually had holes needs an importer at or above this change. Exports of dense uppers are unaffected.

## Measurements

Real snapshot artifact from an agent workload: 4 GiB logical upper with 2,981,888 bytes allocated across 35 extents (macOS, 12 cores, APFS, debug build, 3 runs each; before = this branch with the export change stashed).

|  | before | after |
|---|---|---|
| `Snapshot::export` to `.tar.zst` | 6.44–6.54 s | 0.018–0.096 s |
| archive size | ~2.09 MB | ~7.1 KB |

Round trip through `Snapshot::import`: sha256 over the full 4 GiB logical content matches the source, and the imported upper allocates exactly the same 2,981,888 bytes. A plain-tar export of the same artifact is 2,985,472 bytes (data + headers), and extracting it with system bsdtar yields the same sha256 and the same allocation.

## Tests

New cases in `sdk/rust/tests/snapshot_artifact.rs` (the existing 19 are unchanged and green):

- sparse upper round trip via zstd, with a 512-unaligned extent and a trailing hole: asserts content equality and that the imported upper keeps its holes
- plain-tar export stores only data extents (archive far below the apparent size) and is read back by the sync `tar` crate as an independent reader; the upper's entry type is asserted as `GNUSparse`
- 60+ extents to exercise chained extended sparse headers
- all-hole upper (zero data extents)
- dense upper still produces a `Regular` entry

The sparse tests guard on the test filesystem actually producing sparse files, with the same guard idiom as the `copy.rs` tests. APFS densifies some synthetic mid-file writes, so the synthetic multi-extent cases skip on macOS hosts and run fully on the Linux CI runners; the multi-extent path on macOS is covered by the real-artifact measurements above (35 extents).

## Test plan

- `cargo test -p microsandbox --test snapshot_artifact`: 24 passed, 0 failed
- `cargo clippy -p microsandbox --lib` and `cargo fmt`: clean, no new warnings


appcypher:
this change is important. and there was plan to get this in at some point, so thanks for creating it.

there are some changes i will be making in a follow-up pr to make this support windows as well.

appcypher (approved):

appcypher:
fixing the ci

PR #1173, opened 2026-07-14
feat(go-sdk): support tty exec sessions

kingfs:
## TL;DR

  Add programmable TTY support to the microsandbox Go SDK so Go services can start and control interactive processes inside a sandbox without attaching the host’s local terminal.

  This enables service-side integrations such as agent-compose exec/run -it, where stdin, terminal output, resize events, and signals are transported over an RPC stream.

  ## Motivation

  The Rust SDK already supports interactive exec sessions through:

  - PTY allocation
  - streaming stdin and terminal output
  - terminal resizing
  - signals and process termination

  The Go SDK previously exposed streaming exec, stdin, signals, and process termination, but it could not request a PTY or resize one. Its existing Sandbox.Attach API connects directly to the caller’s local terminal, which works for local CLI applications but not for daemons or remote execution services.

  A Go service needs to control the sandbox terminal programmatically:

  User terminal
    <-> application RPC/WebSocket stream
    <-> Go service
    <-> microsandbox Go SDK
    <-> sandbox PTY
    <-> guest process

  For example, [agent-compose](https://github.com/chaitin/agent-compose) needs this API to implement:
```bash
  agent-compose exec <sandbox> -it bash
  agent-compose run <agent> --command bash -it
```
  The agent-compose daemon does not own the user’s terminal. It must forward stdin bytes, output bytes, resize events, and signals between its bidirectional RPC and the microsandbox exec session.

  ## Description

  This PR brings the Go streaming exec API in line with the existing Rust SDK capabilities.

  ### TTY allocation

  Add WithExecTTY(bool):
  ```go
  handle, err := sandbox.ExecStream(
      ctx,
      "/bin/bash",
      nil,
      microsandbox.WithExecStdinPipe(),
      microsandbox.WithExecTTY(true),
  )
  ```

  The option is passed through:

  Go ExecConfig
    -> Go FFI ExecOptions
    -> JSON FFI contract
    -> native Rust bridge
    -> Rust ExecOptionsBuilder::tty
    -> guest ExecRequest
    -> agentd PTY

  This allocates a real pseudo-terminal inside the guest. It does not merely merge stdout and stderr on the Go side.

  ### Terminal resizing

  Add:
  ```go
  func (h *ExecHandle) Resize(
      ctx context.Context,
      rows, cols uint16,
  ) error
  ```
  The implementation introduces the corresponding msb_exec_resize C ABI function and forwards the operation to the existing Rust ExecControl::resize implementation.

  This allows a Go service to relay terminal resize events such as SIGWINCH-driven changes from a remote client:

  ```go
  if err := handle.Resize(ctx, 40, 120); err != nil {
      return err
  }
  ```

  ### Concurrent exec control

  An interactive service normally has independent input and output loops:

  output goroutine: ExecHandle.Recv()
  input goroutine:  stdin / resize / signal / cancel

  Recv may remain blocked while waiting for guest output. If resize and signal operations require the same mutex as Recv, they cannot reach the guest until output arrives, which can deadlock an interactive session.

  The native bridge now stores each exec session as one lifecycle entry containing:

  - the mutable ExecHandle, used by Recv, Collect, and Wait
  - a cloneable ExecControl, used by Resize, Signal, and Kill

  This keeps registration and cleanup atomic while allowing control operations to proceed independently of a blocked output receive.

  ## API Example

  ```go
  handle, err := sandbox.ExecStream(
      ctx,
      "/bin/bash",
      nil,
      microsandbox.WithExecStdinPipe(),
      microsandbox.WithExecTTY(true),
  )
  if err != nil {
      return err
  }
  defer handle.Close()

  stdin := handle.TakeStdin()
  if stdin == nil {
      return errors.New("stdin pipe is unavailable")
  }

  go func() {
      for {
          event, err := handle.Recv(ctx)
          if err != nil {
              return
          }

          switch event.Kind {
          case microsandbox.ExecEventStdout:
              // In TTY mode, terminal output is emitted through stdout.
              sendTerminalOutput(event.Data)
          case microsandbox.ExecEventExited:
              reportExitCode(event.ExitCode)
          case microsandbox.ExecEventDone:
              return
          }
      }
  }()

  if err := handle.Resize(ctx, 40, 120); err != nil {
      return err
  }

  if _, err := stdin.Write([]byte("echo hello\n")); err != nil {
      return err
  }

  // Closing the sink sends stdin EOF without closing the exec handle.
  return stdin.Close()
  ```
  ## TTY Output Semantics

  When TTY mode is enabled:

  - the guest process runs inside a real PTY
  - stdin is written to the PTY
  - stdout and stderr form one terminal byte stream
  - terminal output is exposed through stdout events
  - stderr cannot be reliably separated, matching normal PTY behavior
  - resize operations update the guest PTY dimensions

  Non-TTY exec behavior remains unchanged and continues to expose separate stdout and stderr events.

  ## Compatibility

  TTY remains disabled by default, so existing callers preserve their current behavior.

  Existing APIs are unchanged:

  - Exec
  - ExecStream
  - TakeStdin
  - Signal
  - Kill
  - Attach

  Sandbox.Attach remains suitable for local CLI applications that want microsandbox to control the caller’s real terminal. WithExecTTY and ExecHandle.Resize provide the programmable alternative for daemons, RPC services, WebSocket servers, and other remote terminal integrations.

  The Go SDK and libmicrosandbox_go_ffi must be upgraded together because this PR adds the msb_exec_resize native symbol.

  ## Testing

  The added coverage verifies:

  - WithExecTTY(true) configures TTY execution
  - the guest command receives a real terminal on stdin and stdout
  - stdin remains writable during a TTY session
  - resize succeeds while another goroutine is blocked in Recv
  - the resized dimensions are visible inside the guest
  - stderr is merged into the terminal stdout stream in TTY mode

  Validation performed:
  ```bash
  cd sdk/go
  go test -count=1 ./...
  go test -race -count=1 ./...

  cargo check -p microsandbox-go
  cargo clippy -p microsandbox-go -- -D warnings
  cargo fmt --all -- --check
  git diff --check
  ```
  The Go unit and race suites pass with 131 tests across 19 packages.

  cargo test -p microsandbox-go reached the final link step but could not complete in the current environment because the system libcap-ng development library was unavailable. The KVM-backed integration test was added but was not run locally because the worktree did not contain the built msb, guest runtime, and Go FFI shared-library artifacts required by the integration harness.


kingfs:
@appcypher @toksdotdev  Please review this pr

toksdotdev:
hey @kingfs. apologies for the delay on this. i should give this a review later today, and should make it to the release going out this wekend 🙏🏽

toksdotdev (approved):
sorry i couldn't get to this on friday as planned.

reviewed, and it lgtm! thanks for the improvements. i'll add an extra commit that documents this function.

PR #1181, opened 2026-07-17
fix(network): propagate guest TCP half-close to proxy tasks

### LINKED ISSUE #1180: Guest FIN never tears down proxied TCP connections: CLOSE_WAIT sockets leak until the 256-slot connection table fills and guest egress goes dark
## Summary

When a guest closes a proxied TCP connection cleanly (FIN — e.g. its process exits while holding an idle HTTP keep-alive), the network stack never tears the connection down. The smoltcp socket sits in `CLOSE_WAIT` forever, the tokio proxy task stays blocked on `from_smoltcp.recv()`, and the upstream host socket stays open (`ESTABLISHED`). Each such connection permanently consumes one slot of the connection table (`DEFAULT_MAX_CONNECTIONS = 256`). Once the table fills, `create_tcp_socket` returns `false` for every new guest SYN, so all new guest connections are refused — the guest sees egress as unreachable while the sandbox itself stays up. Only `msb stop && msb start` recovers.

Connections aborted with RST clean up fine; only the clean-FIN path leaks.

## Environment

- Observed on v0.6.1; mechanism verified unchanged at current `main` (`crates/network/lib/conn.rs` is identical between v0.6.1 and `main`).
- macOS (Apple Silicon) host, libkrun backend, TLS interception enabled — but the defect is in the shared connection tracker beneath both the plain and intercepted paths, so it should apply to any proxied guest TCP.

## Mechanism (file/line refs against current `main`)

All paths relative to `crates/network/`.

1. **The only teardown trigger in the relay loop is proxy-side.** `lib/conn.rs::relay_data` closes the smoltcp socket only when the proxy task has exited (`conn.to_proxy.is_closed()`, ~L305). Nothing inspects the socket's own state (`CloseWait`, `may_recv()`, …) to detect a **guest-initiated** FIN — there is no occurrence of `CloseWait`/`may_recv` anywhere in `lib/`.
2. **Cleanup only reaps fully-`Closed` sockets.** `lib/conn.rs::cleanup_closed` (~L394) removes sockets in `tcp::State::Closed` only. A `CLOSE_WAIT` socket is never `Closed`, so it is never reaped and its `to_proxy` sender is never dropped.
3. **The proxy task therefore never wakes.** Both relay variants exit when `from_smoltcp.recv()` returns `None` (`lib/proxy.rs` plain relay; `lib/tls/proxy.rs::bypass_relay` / `intercept_relay`) — but that only happens when the tracker drops the sender, which (per 1 and 2) never happens after a guest FIN. For an idle keep-alive both `select!` arms are pending forever, so the upstream socket is orphaned in `ESTABLISHED`.
4. **The cycle is circular by construction:** proxy exits ⟸ `from_smoltcp` closes ⟸ sender dropped ⟸ connection reaped ⟸ socket `Closed` ⟸ `socket.close()` called ⟸ proxy exited. A guest FIN enters this cycle at no point.
5. **Slot exhaustion.** Each orphan holds a `HashMap` entry; `create_tcp_socket` (~L225) refuses new sockets at `max_connections` (default 256, `lib/conn.rs` L29). With no socket created for the SYN, the poll loop's auto-RST path means the guest connection can never establish.
6. **Escape hatch = upstream closing first.** If the upstream idle-closes, `server_rx.read()` returns 0, the proxy exits, and the existing `to_proxy.is_closed()` path reaps the socket. So the leak is *permanent* only against upstreams that hold keep-alives open indefinitely (we observed it against `api.anthropic.com`), and self-heals against upstreams that idle-close in seconds (e.g. typical CDNs) — which makes it easy to miss in short tests.

## How we hit it

Long-lived agent workloads inside the sandbox repeatedly kill worker processes that hold idle HTTPS keep-alive connections to an API upstream that never idle-closes. Each kill produces a clean guest FIN and thus one permanently leaked slot; over hours/days the table hits 256 and all guest egress dies, while the VM itself looks healthy. Host-side `ss`/`lsof` show ~256 `ESTABLISHED` upstream sockets held by the msb process; a sandbox restart clears everything.

## Reproduction

Two independent reproductions confirm the mechanism: a deterministic in-process test, and an end-to-end black-box run on a stock released binary.

### In-process (deterministic, no VM needed)

Drive the real `ConnectionTracker` + smoltcp `Interface` through a handshake, hand off the `NewConnection` channels (standing in for an idle proxy task), then inject a guest FIN frame:

- The socket enters `CLOSE_WAIT` and stays there through arbitrarily many `relay_data`/`cleanup_closed` passes (tested out to simulated days).
- The proxy-side `from_smoltcp.try_recv()` stays `Empty` (never `Disconnected`) — i.e. a real proxy task would block forever and never close the upstream.
- Injecting RST instead of FIN tears down cleanly (control).
- A full table refuses new sockets (`create_tcp_socket` → `false`).

Test code exercising exactly this (flipped to assert the fixed behavior) is included in the companion PR.

### End-to-end on a live VM (stock v0.6.6, plain path)

Reproduced black-box against a stock released build (`brew install superradcompany/tap/microsandbox`, `msb 0.6.6`) on the **plain/gateway path** (no `--tls-intercept`) — confirming the leak lives in the shared relay layer, not the TLS-interception path. A host-side "silent listener" (accepts, never reads/writes/closes — modeling a never-idle-closing upstream like `api.anthropic.com`) received idle TCP connections opened by short-lived guest processes that were then `SIGKILL`ed, sending a clean FIN on each idle keep-alive:

- Host-side msb-held `ESTABLISHED` sockets climbed monotonically in lockstep with each successful guest kill, plateauing at **exactly 256** (`DEFAULT_MAX_CONNECTIONS`) over a 285-cycle run.
- At the plateau, new guest connections were refused instantly — both to the listener **and** to public `1.1.1.1:443` (egress globally dead, not port-specific) — while the sandbox VM stayed running and reachable (agent ping ~160 µs; guest `exec`/`uptime` worked).
- `msb stop && msb start` cleared all 256 orphans (msb-held `ESTABLISHED` → 0) and restored egress on a fresh msb pid.
- A pure idle batch of kills accumulated permanently with **no** interleaved live-connection load; slots were freed only by the restart (no reclaim-on-allocation).

This matches the field signature exactly (host socket counts climbing per kill, terminal all-egress-refused state, restart clearing it), so the in-process mechanism and the observed field behavior are the same defect.

## Impact

Any workload that cleanly closes proxied TCP connections to an upstream that doesn't idle-close will eventually kill all guest egress. Interception makes it more visible (long-lived API keep-alives typically flow through the intercepted path), but the defect is in the shared tracker, not the TLS layer.

## Suggested fix

Fix offered in: #1181

brandonduff:
Fixes: #1180 
## TL;DR

A guest-initiated FIN on a proxied TCP connection was never propagated, leaking one connection-table slot per clean close until the 256-slot table filled and all guest egress went dark (recoverable only by a sandbox restart). This propagates the guest half-close to the proxy task so the upstream write half is shut down and the slot is reclaimed, without breaking legal TCP half-close.

## Description

- **Root cause:** the connection tracker only tore connections down from the proxy side (`to_proxy.is_closed()`), and `cleanup_closed` only reaps fully-`Closed` sockets. A guest FIN left the smoltcp socket in `CLOSE_WAIT` forever, the proxy task blocked on `from_smoltcp.recv()`, and the upstream socket open — permanently consuming one of the `max_connections` (default 256) slots whenever the upstream doesn't idle-close. Once the table filled, every new guest SYN was refused and guest egress went dark until the sandbox was restarted. (Full mechanism: #<issue>.)
- **`conn.rs` (shared tracker):** `Connection::to_proxy` becomes `Option<Sender>`. When the socket is in `CLOSE_WAIT` and everything the guest sent has been relayed (`read_buf` flushed, receive buffer drained), the sender is dropped, so the proxy task's `from_smoltcp.recv()` returns `None`. The server → guest direction keeps relaying. Proxy-task exit — which closes and reaps the socket via the existing path — is detected via `from_proxy.is_closed()` once the sender is gone.
- **Relay loops (`proxy.rs` plain + CONNECT tunnel, `tls/proxy.rs` bypass + intercept):** `from_smoltcp` EOF no longer breaks the loop. Instead the upstream **write** half is shut down (TCP FIN; for intercepted TLS, `close_notify` + FIN) and the server → guest direction keeps relaying until the upstream closes. A TLS 1.3 upstream may keep sending after `close_notify`; a TLS 1.2 upstream answers with its own `close_notify`, ending the relay — either way the task exits and the socket is reaped.
- **Why half-close, not full teardown on `CLOSE_WAIT`:** a guest FIN with response data still in flight is legal TCP (`shutdown(SHUT_WR)` then read). Closing on `CLOSE_WAIT` would be a smaller (tracker-only) diff but would drop the server's response to a half-closing guest — a regression. Half-close propagation matches what kernel NAT / userspace slirp-style stacks do, preserves request-FIN-response patterns, and still reclaims the slot.
- **End-to-end lifecycle after this change:** guest FIN → tracker drops sender → proxy shuts down upstream write → upstream finishes/closes → proxy exits → tracker closes socket (`CLOSE_WAIT` → `LAST_ACK` → `Closed`) → slot reclaimed.
- **Residual (pre-existing) gap:** an upstream that neither sends data nor closes after receiving our FIN can still pin a slot; on this build such a socket sits in `FIN_WAIT_2` (the FIN was sent) rather than `ESTABLISHED`. A `CLOSE_WAIT`/idle timeout on guest TCP sockets (analogous to `UDP_PEER_TIMEOUT`) would close it deterministically; kept out of this PR to stay minimal.

## Test Plan

New regression tests in `stack.rs` drive the real `ConnectionTracker` + smoltcp `Interface` through handshake and teardown with hand-built frames:

- `guest_fin_propagates_half_close_without_killing_the_connection` — FIN ⇒ proxy channel sees EOF (`Disconnected`), connection stays tracked, and pending server → guest data is still delivered to the half-closed guest. Fails without this fix (channel stays `Empty` forever — the orphan).
- `guest_fin_connection_is_reaped_after_proxy_exit` — after the proxy exits, the guest gets our FIN, its final ACK completes the close, and the slot is reclaimed.
- `guest_rst_is_cleaned_up` — RST teardown stays clean.
- `full_connection_table_refuses_new_sockets` — table-limit gating.

Commands:

- [x] `cargo test -p microsandbox-network --lib` — 415 passed, 0 failed (includes the four regression tests above)
- [x] `cargo clippy -p microsandbox-network --lib --tests` — clean
- [x] `cargo fmt --check` — clean
- [x] Patch applies cleanly to v0.6.1, where the bug was first observed (`conn.rs` unchanged between v0.6.1 and `main`)

Black-box validation on a live VM (stock v0.6.6 vs a build carrying this fix, plain/gateway path, 285 guest-kill cycles per cell):

- [x] Stock + FIN-respecting upstream — leaks: msb-held `ESTABLISHED` plateaus at 256, guest egress dies, `msb stop && msb start` heals; 0 upstream EOF-closes during the run (the FIN is never propagated).
- [x] Fixed + FIN-respecting upstream — no accumulation: max msb-held = 1 across 285 kills, `ok=285 fail=0`, upstream accepts:EOF-closes 1:1, egress healthy throughout. Field-realistic case (the `api.anthropic.com` profile) — the pass signal.
- [x] Fixed + truly-silent upstream — still pins slots (the residual gap above): plateaus at 256 in `FIN_WAIT_2` (vs `ESTABLISHED` on stock), egress dies, heal → 0. Expected-by-design; motivates the idle-timeout follow-up.
- [x] Half-close preserved — a guest that `shutdown(SHUT_WR)`s then reads still receives a server payload sent after the FIN; no premature teardown.

### DIFF
--- a/crates/network/lib/conn.rs
+++ b/crates/network/lib/conn.rs
@@ -90,7 +90,12 @@ struct Connection {
     /// Original destination (from the guest's SYN).
     dst: SocketAddr,
     /// Sends data from smoltcp socket to proxy task (guest → server).
-    to_proxy: mpsc::Sender<Bytes>,
+    ///
+    /// Set to `None` once the guest half-closes (FIN) and all its data has
+    /// been relayed: dropping the sender makes the proxy task's
+    /// `from_smoltcp.recv()` return `None`, propagating the half-close
+    /// upstream while the server → guest direction stays open.
+    to_proxy: Option<mpsc::Sender<Bytes>>,
     /// Receives data from proxy task to write to smoltcp socket (server → guest).
     from_proxy: mpsc::Receiver<Bytes>,
     /// Proxy-side channel ends, held until the connection is ESTABLISHED.
@@ -257,7 +262,7 @@ impl ConnectionTracker {
             Connection {
                 src,
                 dst,
-                to_proxy: to_proxy_tx,
+                to_proxy: Some(to_proxy_tx),
                 from_proxy: from_proxy_rx,
                 proxy_channels: Some(ProxyChannels {
                     from_smoltcp: to_proxy_rx,
@@ -302,7 +307,14 @@ impl ConnectionTracker {
             // an RST via `abort()` is instead sent so happy-eyeballs
             // clients fall back to another family instead of committing
             // to this half-open connection.
-            if conn.to_proxy.is_closed() {
+            let proxy_exited = match &conn.to_proxy {
+                Some(to_proxy) => to_proxy.is_closed(),
+                // The guest already half-closed (sender dropped below), so
+                // proxy exit is detected on the other channel instead: the
+                // proxy drops its `to_smoltcp` sender when it returns.
+                None => conn.from_proxy.is_closed(),
+            };
+            if proxy_exited {
                 if matches!(
                     conn.proxy_connect.status(),
                     ProxyConnectStatus::UpstreamConnectFailed
@@ -330,24 +342,39 @@ impl ConnectionTracker {
             }
 
             // smoltcp → proxy: flush read_buf first, then read from socket.
-            if let Some(pending) = conn.read_buf.take()
-                && let Err(e) = conn.to_proxy.try_send(pending)
-            {
-                conn.read_buf = Some(e.into_inner());
-            }
+            if let Some(to_proxy) = &conn.to_proxy {
+                if let Some(pending) = conn.read_buf.take()
+                    && let Err(e) = to_proxy.try_send(pending)
+                {
+                    conn.read_buf = Some(e.into_inner());
+                }
 
-            if conn.read_buf.is_none() {
-                while socket.can_recv() {
-                    match socket.recv_slice(&mut relay_buf) {
-                        Ok(n) if n > 0 => {
-                            let data = Bytes::copy_from_slice(&relay_buf[..n]);
-                            if let Err(e) = conn.to_proxy.try_send(data) {
-                                conn.read_buf = Some(e.into_inner());
-                                break;
+                if conn.read_buf.is_none() {
+                    while socket.can_recv() {
+                        match socket.recv_slice(&mut relay_buf) {
+                            Ok(n) if n > 0 => {
+                                let data = Bytes::copy_from_slice(&relay_buf[..n]);
+                                if let Err(e) = to_proxy.try_send(data) {
+                                    conn.read_buf = Some(e.into_inner());
+                                    break;
+                                }
                             }
+                            _ => break,
                         }
-                        _ => break,
                     }
+                }
+
+                // Guest half-close: the guest sent a FIN (CLOSE_WAIT) and
+                // everything it sent has been relayed. Drop the sender so
+                // the proxy task sees EOF and can shut down the guest →
+                // server direction upstream. The server → guest direction
+                // stays open; the socket is closed once the proxy task
+                // exits (see `proxy_exited` above).
+                if matches!(socket.state(), tcp::State::CloseWait)
+                    && conn.read_buf.is_none()
+                    && !socket.can_recv()
+                {
+                    conn.to_proxy = None;
                 }
             }
 
--- a/crates/network/lib/proxy.rs
+++ b/crates/network/lib/proxy.rs
@@ -355,10 +355,11 @@ async fn tcp_proxy_task(
     //
     // guest → server: receive from channel, write to server socket.
     // server → guest: read from server socket, send via channel + wake poll.
+    let mut guest_eof = false;
     loop {
         tokio::select! {
             // Guest → server: substitute placeholders before forwarding.
-            data = from_smoltcp.recv() => {
+            data = from_smoltcp.recv(), if !guest_eof => {
                 match data {
                     Some(bytes) => {
                         if let Some(tls_state) = late_connect_state.take()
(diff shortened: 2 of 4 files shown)


toksdotdev (approved):
hey, thanks for the fix. very much appreciated. 🥳

building upon this work, i also opened #1255 to address other half-close issues i discovered from a broader code sweep.

Issue #1201, opened 2026-07-24
agentd PID 1 does not reap orphaned children, causing persistent zombie accumulation


Fishbiscuit:
### Summary

In a long-running Microsandbox guest, exited orphan descendants remain as PID-1-owned zombies indefinitely. A live development sandbox accumulated hundreds of them, primarily `git`, and the count continued growing during investigation.

### Reproduction Steps

```bash
msb create --name zombie-repro ubuntu

# Launch a child that outlives its direct exec parent.
msb exec zombie-repro -- sh -c 'sleep 0.1 &'
sleep 1

# Inspect PID 1 and its defunct children.
msb exec zombie-repro -- sh -c 'ps -p 1 -o pid,ppid,stat,comm,args; ps -eo pid,ppid,state,comm'

# Repeat the orphaning exec and watch this count grow.
msb exec zombie-repro -- sh -c "ps -e -o state= | grep -c '^Z'"
```

Repeat the background-child command followed by the count command. The count increases rather than returning to zero.

### Expected Behavior

When a descendant is reparented to guest PID 1 and later exits, `agentd` should reap it. A persistent sandbox should not accumulate zombies across exec sessions.

### Actual Behavior

In the observed guest:

```text
PID  PPID STAT COMMAND    COMMAND
1       0 Sl   init.krun  /init.krun
```

At 10:00 UTC, `/proc` contained 793 zombies out of 950 processes, all with PPID 1:

```text
772 git
14 sleep
4 python3
3 python
```

A later sample at 10:01 UTC contained 816 zombies out of 978 processes, demonstrating continued accumulation. The guest has `kernel.pid_max = 8192`, so this steadily consumes a finite PID space. Of the initial 772 zombie `git` processes, 748 had exit status 0 and 24 had exit status 1: these processes had terminated, but were not reaped.

Stopping/removing and recreating the sandbox clears the table; normal command completion does not.

### Root Cause Analysis

**Source:** live `/proc` inspection plus static inspection of `agentd` v0.5.3, v0.6.6, and current `main`

[...]

PR #1212, opened 2026-07-26
fix(agentd): centralize child process reaping

crpboy:
Give agentd a process-wide child-status owner so concurrent exec sessions cannot lose fast exits to competing waiters, while PID 1 also reaps adopted descendants instead of accumulating zombies.

- Start ProcessManager once before the agent loop and keep child-status ownership on a dedicated SIGCHLD thread. Each signal drains waitpid(-1, WNOHANG), resolves registered exit watchers, and also consumes untracked adopted children.
- Guard the OS spawn-to-PID-registration window with a shared RwLock read guard. Spawns remain concurrent, while an exclusive reap pass cannot observe a fast exit before its PID is present in the registry; terminal manager failures reject new spawns and wake existing watchers.
- Replace per-session waitpid and Tokio Child ownership with ProcessExitWatcher for both pipe and PTY execution. Pipe creation uses std::process::Command so Tokio's process driver cannot race the central manager, and both output paths drain stdout/stderr before publishing ExecExited.
- Preserve existing wire behavior for signal termination, keep spawn failures recoverable, and leave teardown's native waitpid sweep as the final fallback when the process manager can no longer be assumed healthy.
- Add unit and real-microVM coverage for concurrent rapid exits, adopted descendants, pipe and PTY status delivery, signal termination, large interleaved output, spawn failures, runtime replacement, and recovery after failure.

- [x] cargo fmt --all -- --check
- [x] cargo clippy --manifest-path crates/agentd/Cargo.toml --all-targets -- -D warnings
- [x] cargo test --manifest-path crates/agentd/Cargo.toml (122 passed)
- [x] cargo build --release --manifest-path crates/agentd/Cargo.toml --target x86_64-unknown-linux-musl
- [x] cargo test -p microsandbox --test child_reaping -- --ignored --nocapture (1 passed on a real KVM microVM)


crpboy:
Fixes #1201

appcypher:
@crpboy thanks for taking this on. child reaping gets tricky quickly, and i think the main idea is right: we need one owner for child exit statuses, otherwise an orphan reaper could race the pipe and pty waiters.

there are two things i think we should address before merging:

1. **pid reuse when signalling**

   the manager reaps a child immediately, but its session remains signalable until output has drained. the pid/process-group id could be reused during that window, so a late signal might reach an unrelated command.

https://github.com/superradcompany/microsandbox/blob/a8699444df58fa14d01df26fd4f36f89ae755cbb/crates/agentd/lib/process.rs#L256-L264

https://github.com/superradcompany/microsandbox/blob/a8699444df58fa14d01df26fd4f36f89ae755cbb/crates/agentd/lib/session.rs#L344-L357

   signalling should go through the manager using a unique registration id, or become disabled atomically when the child is reaped.

2. **the unbounded reap loop**

   the manager holds the spawn gate while draining `waitpid`. a workload continuously producing exited children could block new execs. could we reap in bounded batches and schedule another pass when needed?

https://github.com/superradcompany/microsandbox/blob/a8699444df58fa14d01df26fd4f36f89ae755cbb/crates/agentd/lib/process.rs#L217-L226

https://github.com/superradcompany/microsandbox/blob/a8699444df58fa14d01df26fd4f36f89ae755cbb/crates/agentd/lib/process.rs#L256-L278

i’d also simplify where possible. the dedicated thread is reasonable, but the global singleton, multiple locks, thread-local guard, and runtime-replacement support may be more than we currently need. creating the manager once, passing it down, and consolidating its state should be easier to maintain.

the test coverage is appreciated. there are a few broader lifecycle gaps i’ll handle separately, so you don’t need to cover everything in this pr. overall, the central-reaper direction is good. i mainly want us to tighten process identity handling and keep it as simple as we can.

appcypher:
if you are busy, i can implement the fixes on top of what you have already since you have maintainer edits available.

i think this fix is important to ship before the next release bump.

crpboy:
> if you are busy, i can implement the fixes on top of what you have already since you have maintainer edits available.
> 
> i think this fix is important to ship before the next release bump.

@appcypher Apologies for the late reply, I’ve been a bit swamped lately :(  Would appreciate it if you could take this forward.

PR #1227, opened 2026-07-28
feat(network): constrain secrets to exact request headers

lirazsiri:
## TL;DR

Adds provider-neutral exact-header secret placement so a host-held secret can be substituted only into one configured initial HTTP request header and optional authentication scheme.

## Description

- add exact-header injection rules for HTTP/1.1 and HTTP/2, including strict handling of malformed values, duplicate target fields, trailers, and reused HTTP/2 streams
- expose create-time exact-header placement through the CLI and Rust, Python, TypeScript, and Go SDKs
- make convenience builders disable broad header and Basic Auth injection when exact-header placement is selected
- preserve legacy `--secret` behavior and serialized defaults
- update SDK references, security documentation, and runnable examples

## Test Plan

- `cargo fmt --all -- --check`
- `cargo test -p microsandbox-network exact_header` — 2 passed after rebasing onto current upstream `main`
- `cargo test -p microsandbox-types secret_exact_header` — 1 passed after rebase
- `cargo test -p microsandbox-cli secret_exact_header` — 2 passed before the conflict-free rebase
- Node unit suite — 109 passed before rebase
- Node TypeScript typecheck — passed before rebase
- external real-runtime fake-API proof — exact `Authorization: Bearer` placement passed before and after live rotation; wrong header, missing/wrong scheme, combined value, duplicate fields, and a disallowed authority received only placeholders

Post-rebase CLI retesting is currently blocked by the upstream Rust SDK build script requesting `libkrunfw.so.5.6.1`, which is absent from the downloaded/local runtime bundle. The exact-header patches were unchanged by the rebase.

### DIFF
--- a/crates/cli/lib/commands/common.rs
+++ b/crates/cli/lib/commands/common.rs
@@ -555,6 +555,19 @@ pub struct SandboxOpts {
     #[arg(long)]
     pub secret: Vec<String>,
 
+    /// Inject a secret only as one exact request-header credential.
+    ///
+    /// Format:
+    /// `ENV@HOST[,HOST...];header=NAME[;scheme=SCHEME]`.
+    /// The optional scheme is followed by exactly one space before the
+    /// placeholder. This option is available only while creating a sandbox.
+    #[cfg(feature = "net")]
+    #[arg(
+        long = "secret-exact-header",
+        value_name = "ENV@HOST;header=NAME[;scheme=SCHEME]"
+    )]
+    pub secret_exact_header: Vec<String>,
+
     /// Action when a secret is sent to a disallowed host (block, block-and-log, block-and-terminate, passthrough).
     #[cfg(feature = "net")]
     #[arg(long)]
@@ -690,6 +703,7 @@ impl SandboxOpts {
             || !self.tls_upstream_ca_cert_for.is_empty()
             || !self.tls_no_verify_upstream_for.is_empty()
             || !self.secret.is_empty()
+            || !self.secret_exact_header.is_empty()
             || self.on_secret_violation.is_some();
 
         #[cfg(not(feature = "net"))]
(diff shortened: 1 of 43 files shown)


toksdotdev:
hey @lirazsiri , thanks for the pr. do you mind sharing what your use case is?

lirazsiri:
Our use case is running untrusted workloads while keeping provider credentials outside the guest. The guest receives a placeholder, and the host-side proxy substitutes the real credential only when an outbound request satisfies the configured policy.

Host allow-listing alone is not a sufficient release condition. With broad header substitution, an untrusted guest can place the placeholder in any request header sent to an allowed authority. That creates several credential-disclosure paths:

- A diagnostic, echo, upload, or error endpoint may reflect arbitrary headers.
- Metadata headers may be logged, forwarded, copied into tracing systems, or retained by intermediaries.
- Duplicate authorization fields, trailers, or ambiguous HTTP/1 and HTTP/2 placement can cause components to disagree about where the credential was supplied.
- A placeholder embedded in a larger or unexpected value can materialize the credential outside the intended authentication form.

Exact-header placement makes credential release depend on both destination and syntax. For example, the host can authorize only:

```text
Authorization: Bearer <placeholder>
```

The proxy substitutes only the value of that exact initial request header with the configured scheme. It rejects duplicate target headers, trailers, malformed or combined values, and reused HTTP/2 streams where placement is ambiguous.

This does not prevent an authorized request to the allowed API. It prevents the plaintext credential from being materialized in arbitrary guest-selected header locations, reducing reflection and forwarding opportunities while providing a meaningful least-authority policy.

The feature is provider-neutral because the same requirement applies to `Authorization: Bearer`, `Proxy-Authorization`, and provider-specific API-key headers. It is opt-in, and existing broad secret behavior remains compatible.

lirazsiri:
Status update: the PR branch has now been rebased onto current upstream `main` at `288ef7c8` (v0.6.16) and force-updated with a lease. GitHub reports the new head as `21b069f2`; the previous conflict state is resolved. The focused network, types, and CLI tests pass, and the rebuilt binary passed complete isolated `msb` and LibreSBX lifecycle checks through create, list, inspect, exec, stop, restart, re-exec, remove, and final cleanup.

Could the maintainers provide a progress report on review and any remaining blockers? If #1356 is intended to supersede this PR, please clarify how it will preserve exact request-header placement and address this threat.

This remains an important credential-security boundary for untrusted workloads. Host allow-listing alone lets a guest place the placeholder in an arbitrary header sent to an allowed authority. Diagnostic or echo endpoints, and downstream logging or forwarding systems, can then reflect or retain the materialized token. Without pinning substitution to one exact initial request header and optional scheme, it is difficult to secure credentials against token-reflection attacks.

toksdotdev:
hey @lirazsiri, my honest apologies for the delay on this. had a lot going on our end, and finally catching up with PRs.

when you've got some time, please could you help rebase? a lot has changed ... specifically, we've made the secret component a lot more flexible, so it does overlap a lot with some of the changes you've made.

thanks a lot.

PR #1231, opened 2026-07-28
fix(sdk/node-ts): avoid TS reserved word as napi parameter name (CI-breaking under napi-rs cli 3.8.0)

nuri-yoo:
## TL;DR

Every `Check (linux-x86_64 / linux-aarch64 / darwin-aarch64)` job now fails in the Build Node SDK step because `@napi-rs/cli` 3.8.0 (released 2026-07-28 09:19 UTC) validates the generated `index.d.ts` and rejects it: `SecretBuilder.env` takes its argument as `var`, which is not a legal parameter name in TypeScript. Renaming the Rust parameter to `name` fixes the generated declaration with no JavaScript API change. This unblocks CI for every open PR.

## What broke

Any CI run that reaches the Build Node SDK step of `check.yml` after 2026-07-28 09:19 UTC fails with:

```
Internal Error: Failed to parse declaration source:
1318:7 'var' is not allowed as a parameter name.
1318:18 Expected '=' for property initializer.
1318:19 Unexpected token. A constructor, method, accessor, or property was expected.
    at throwDeclarationDiagnostics (.../@napi-rs/cli/dist/cli.js:3607:8)
    at createDeclarationProgram (.../@napi-rs/cli/dist/cli.js:3594:30)
    at rewriteTypeImportReferences (.../@napi-rs/cli/dist/cli.js:3451:34)
    at generateTypeDef (.../@napi-rs/cli/dist/cli.js:7254:29)
```

First observed on PR #1228 (run [30347794298](https://github.com/superradcompany/microsandbox/actions/runs/30347794298)), whose diff touches only `crates/image`, `crates/network`, and Cargo metadata, so the failure is independent of any PR content. `Check Windows` and the `Test` workflow keep passing because they never build the Node SDK.

## Root cause

The Build Node SDK step installs with `npm install --package-lock=false`, and `sdk/node-ts/package.json` declares `"@napi-rs/cli": "^3"`, so every CI run resolves the newest 3.x release. The timeline:

| Event | Time (UTC) | CLI version resolved |
| --- | --- | --- |
| Last green `Check` on `main` (#1222 merge) | 2026-07-27 19:02 | 3.7.4 |
| `@napi-rs/cli` 3.8.0 published to npm | 2026-07-28 09:19 | — |
| First failing run (PR #1228) | 2026-07-28 09:43 | 3.8.0 |

3.8.0 added a post-generation step (`rewriteTypeImportReferences`) that loads the emitted `.d.ts` into a TypeScript program and throws on parse diagnostics. The declaration it rejects comes from `sdk/node-ts/native/secret_builder.rs`:

```rust
#[napi]
pub fn env(&mut self, var: String) -> &Self {
```

napi-rs copies Rust parameter names verbatim into the declaration, producing `env(var: string): this`. `var` is a reserved word in TypeScript parameter position, so the generated file has never been valid TypeScript. The copy of `native/index.d.ts` committed on `main` carries the same defect at line 1309; running plain `tsc --noEmit` against it reports exactly the errors CI now surfaces:

```
index.d.ts(1309,7): error TS1390: 'var' is not allowed as a parameter name.
index.d.ts(1309,18): error TS1442: Expected '=' for property initializer.
index.d.ts(1309,19): error TS1068: Unexpected token. A constructor, method, accessor, or property was expected.
```

Versions up to 3.7.4 emitted the file without checking it, and the SDK's own `tsconfig.json` sets `skipLibCheck: true`, so nothing caught it earlier. 3.8.0 turned the latent defect into a hard build failure.

## The fix

Rename the Rust parameter from `var` to `name`. The parameter is positional in both Rust and the generated JavaScript binding, so nothing about the runtime or the JS call surface changes; only the generated signature becomes `env(name: string): this`, which is valid TypeScript. `name` matches what the docstring describes ("environment variable name exposed to the sandbox") and the existing test call sites (`.env("API_KEY")`) are unaffected.

I swept every file in `sdk/node-ts/native/` for other TypeScript reserved words in parameter position (`var`, `let`, `const`, `class`, `function`, `enum`, `new`, `delete`, `typeof`, `this`, `super`, `return`, `yield`, `static`, `import`, `interface`, `private`, `protected`, `public`, and the rest of the reserved set). This was the only occurrence.

### Why not pin the CLI instead

Pinning `@napi-rs/cli` to 3.7.4 would turn CI green again but would keep publishing a type declaration that is not parseable TypeScript, and would leave the repository unable to move to any future CLI release until the parameter is renamed anyway. The rename fixes the declaration itself and builds cleanly under both 3.7.4 and 3.8.0. Whether the workspace additionally wants a lockfile-based install for the SDK build is a separate CI-policy question I left untouched.

### Note on the regenerated files

`native/index.d.ts` and `native/index.cjs` are committed build outputs, so this PR commits the files exactly as `napi build` regenerates them. Beyond the one-line `env(var → name)` signature fix, the diff includes codegen updates that are new in `@napi-rs/cli` 3.8.0 and unrelated to this bug: an explicit `__NapiRsAsyncGenerator` helper type for `Symbol.asyncIterator` signatures, and loader hardening in `index.cjs` (error-chain construction and stricter `NAPI_RS_FORCE_WASI` handling). Any fresh checkout that runs the build now produces this output, since the install floats within `^3`.

## Test plan

- [x] Reproduced the failure locally before the fix: `npm install --package-lock=false` resolved `@napi-rs/cli@3.8.0` and `npm run build:native:debug` failed with the identical `Internal Error` at `1318:7`, matching the CI logs line for line.
- [x] `npm run build:native:debug` after the fix: completes, `index.d.ts` line 1318 now reads `env(name: string): this`.
- [x] `npm run build:ts` and `npm run typecheck`: both exit 0.
- [x] `npm run test:unit`: 8 files, 108 tests passed, including the `SecretBuilder` tests that call `.env("API_KEY")`.
- [x] `cargo fmt -p microsandbox-node -- --check` and `cargo clippy -p microsandbox-node --all-targets -- -D warnings`: clean.
- [x] Verified the committed `main` copy of `index.d.ts` fails plain `tsc --noEmit` with TS1390/TS1442/TS1068 at line 1309, and the regenerated copy no longer produces any syntax diagnostic.


nuri-yoo:
Superseded by #1232. Closing.

PR #1243, opened 2026-07-30
fix(network): fail over to subsequent DNS upstreams

### LINKED ISSUE #1229: DNS forwarder does not fail over to subsequent upstream nameservers
## Environment

- Microsandbox: 0.6.7
- Observed on Windows 11 x86-64
- The reproduction does not depend on Windows

## Problem

The DNS forwarder accepts or discovers multiple upstream nameservers,
but only tries the first one. If that server is unusable, the guest
cannot resolve names even when a later server works.

## Minimal reproduction

This reproduction does not modify the host's network configuration or
depend on its existing DNS servers. `192.0.2.1` is from TEST-NET-1,
which is reserved for documentation and should not provide DNS.

```text
msb run alpine --no-tty --dns-nameserver 192.0.2.1 --dns-nameserver 1.1.1.1 --dns-query-timeout-ms 1000 -- /bin/sh -lc "nslookup example.com"
```

The lookup fails because only `192.0.2.1` is attempted.

Reverse the server order:

```text
msb run alpine --no-tty --dns-nameserver 1.1.1.1 --dns-nameserver 192.0.2.1 --dns-query-timeout-ms 1000 -- /bin/sh -lc "nslookup example.com"
```

The lookup succeeds, demonstrating that the second server in the first
command was usable but never attempted.

## Expected behavior

When multiple upstream nameservers are configured or discovered, the
forwarder should try a subsequent server after timeout or transport
failure.

## Actual behavior

The forwarder uses only the first upstream. The guest receives the
Microsandbox gateway as its only nameserver, so it cannot try the
remaining host upstreams itself.

## Suspected cause

`DnsForwarder::build` subsequently uses only `upstreams[0]`.

The comment there says the guest stub can fall through to its next
nameserver, but the guest only knows the Microsandbox gateway. The
remaining host DNS servers are never exposed to the guest or attempted
by the forwarder.

Relevant files:

- `crates/network/lib/dns/forwarder.rs`

## How Windows exposed it

This was discovered through automatic DNS configuration on Windows 11.
The Windows implementation added by #1019 collects DNS servers from
every active adapter. A resolver that Windows itself does not select can
therefore become the forwarder's first and only upstream.

That discovery path made the problem visible, but the explicit
two-server reproduction shows that failover itself is platform
independent.

## Possible fix

Retain the configured or discovered upstream list and attempt subsequent
servers after timeout or transport failure.

MannXo:
Fixes #1229.

## Summary

`DnsForwarder::build` resolved the full upstream list and then kept only `upstreams[0]`, so an unusable first nameserver made every lookup fail even when a later one worked.

The comment at the discard site explained this as safe because the guest's stub would "fall through to its own next nameserver". It cannot: the guest is handed the Microsandbox gateway as its only nameserver, so there is nothing behind it to fall through to. Nothing else in the path was going to try the remaining servers, which is why the reporter's `192.0.2.1` then `1.1.1.1` case failed while the reverse order succeeded.

The forwarder now keeps every upstream and tries them in order, moving on after a per-query timeout or a transport failure.

## Changes

- `DnsForwarder` holds `Vec<ConfiguredUpstream>` instead of a single client plus address. Each entry owns its address, a UDP client and a lazy TCP client.
- `forward_to_configured` walks the list and stops at the first upstream that answers.
- `send_query` isolates one upstream attempt, so "this server is unusable" is a single, explicit condition rather than two early returns.
- Removed the incorrect comment about guest-side fallthrough.
- Documented the ordering and its cost in `docs/networking/dns.mdx`.

Pre-building a UDP client per upstream is free: `build_udp_client` notes that since hickory 0.26 the constructor only wraps a request sender, so socket errors surface per query rather than at build time. TCP stays lazy, and is now built per upstream only when a query actually reaches that upstream, preserving the existing "many sandboxes never use TCP DNS, so don't pay the handshake up front" property.

## Deliberate boundaries

Three decisions that are easier to review as decisions than to infer from the diff.

**Only the configured path fails over.** A query the guest aimed at a specific resolver via `@target` still goes to exactly that resolver. Substituting a different server there would silently redirect a query the guest addressed deliberately, which is a different behavior change from the one this issue asks for.

**A response that arrives is an answer, including `SERVFAIL` and `REFUSED`.** Those do not trigger a retry against the next server. The issue scopes failover to timeout and transport failure, and re-asking on an rcode would change what the sandbox resolves rather than repairing reachability.

**An upstream whose client cannot be built is skipped, not fatal.** `build` only gives up when no upstream is usable at all. Previously a single failure at index 0 took down the whole forwarder.

## Behavior changes worth calling out

Both follow from the fix rather than being incidental, and `AGENTS.md` asks for regressions and material changes to be surfaced:

- **Latency on a cold miss.** Each unreachable upstream now costs up to `query_timeout` before the next is tried, so the worst case is roughly `N x query_timeout` instead of a single timeout. With the issue's `--dns-query-timeout-ms 1000` and three discovered servers that is up to ~3s. I chose to try the whole list rather than cap it, since resolver lists are short and the alternative is failing a name that is resolvable, but a cap is easy to add if you would rather bound it.
- **A name the first resolver could not answer may now be resolved by a later one.** That is the requested behavior, but it does widen what a sandbox can resolve when the host's list is heterogeneous, which is exactly the Windows multi-adapter situation from the issue.

## Testing

Four new tests in `crates/network/lib/dns/forwarder.rs`. They are hermetic: two loopback UDP servers, no external network, no microVM, no host resolver. This mirrors the black-hole-socket approach already used by `dns::client::tests::udp_upstream_honors_query_timeout`.

| Test | Asserts |
| --- | --- |
| `configured_upstream_falls_over_to_the_next_on_timeout` | the reported bug: a stalled first upstream falls over, and the working one is queried exactly once |
| `first_usable_upstream_answers_without_consulting_the_rest` | the reversed order still works, the answer comes from the first server, and later servers are not consulted |
| `configured_upstreams_fall_over_past_several_dead_servers` | failover walks the whole list, not just one extra server |
| `all_upstreams_unusable_yields_servfail` | every upstream dead gives a definite SERVFAIL with the guest's transaction id preserved, not a hang |

Every assertion was sabotage-verified rather than only run green:

| Sabotage | Result |
| --- | --- |
| take only the first upstream (the pre-fix behavior) | the two failover tests **FAIL**, the two guards pass |
| drop the early `break` so every upstream is queried | `first_usable_upstream_answers_without_consulting_the_rest` **FAILS** |

The second sabotage is worth mentioning because it caught a weakness in my own test. The no-fan-out test originally paired a live first server with a black-hole second one, and passed even when the forwarder queried both, since a silent server's reply never arrives and the observable result is identical. The second upstream is now a live responder with a distinct address and a hit counter, so fan-out is actually detectable.

Commands:

```
cargo test -p microsandbox-network       427 passed; 0 failed   (423 on this branch point, +4 new)
cargo test --workspace                   2337 passed; 0 failed
cargo clippy -p microsandbox-network --all-targets -- -D warnings    clean
cargo fmt --all -- --check               clean
```

One note on `cargo clippy --workspace -- -D warnings`: it reports 4 pre-existing errors outside this change, in `packages/agent-client`, `crates/protocol` and `crates/filesystem` (`items after a test module`, `unnecessary_cast`). They reproduce with my diff stashed, and the repo pins no toolchain while CI tracks `stable`, so they look like newer-lint artifacts on rustc 1.97.1 rather than anything on this branch. I left them alone; happy to fix them in a separate PR if useful.


toksdotdev (approved):
hey, really appreciate the changes. lgtm!

PR #1246, opened 2026-07-30
feat(exec): add combined stdout/stderr output option

### LINKED ISSUE #723: Add an option to exec to receive combined stdout/stderr output
Some programs use both stderr and stdout at the same time and the order of outputs between the two matters (Mostly more interactive stuff), to be able to preserve the order in such a case it is necessary to redirect stderr to stdout so that they use one output stream. (`2>&1`, in the shell), [`subprocess.STDOUT`](https://docs.python.org/3/library/subprocess.html#subprocess.STDOUT), or setting `stderr` to `1` using Node.js `child_process` `stdio` option.

ya-luotao:
Closes #723

## What

Adds an opt-in combined-output mode to `exec` that redirects the command's stderr into its stdout stream (like `2>&1`), preserving the true interleaved order of writes — across the guest agent, the wire protocol, all four SDKs, and the CLI:

| Surface | API |
| --- | --- |
| Rust SDK | `ExecOptionsBuilder::combined_output(true)` |
| Python SDK | `combined_output=True` on `exec` / `shell` / `exec_stream` / `shell_stream` |
| node-ts SDK | `.combinedOutput(true)` |
| Go SDK | `WithExecCombinedOutput()` / `ExecConfig.CombinedOutput` |
| CLI | `msb exec --combined-output` |

## How

The merge happens at the only layer that can preserve emission order: the guest process's own file descriptors. In pipe mode, agentd points the child's stderr at the stdout pipe via `dup2(1, 2)` in a `pre_exec` hook — the standard library has already dup2'd the pipes onto fds 0–2 by that point, so the redirect lands stderr writes in the stdout pipe and the kernel guarantees ordering on the shared pipe. No stderr pipe is created (`Stdio::null()`), and no userspace re-ordering is involved. In TTY mode the flag is a no-op: a PTY is inherently a single combined stream.

On the wire, `ExecRequest` gains an optional `#[serde(default)] combined_output` field. The host relay routes frames without parsing CBOR, so no host-side changes are needed.

## Compatibility

- Old peer → new agentd: missing field decodes as `false` (covered by a protocol test).
- New SDK → old agentd: the flag is silently ignored and streams stay separate — no data loss, exactly the degradation VERSIONING.md promises for this class of change. Verified empirically against a stale prebuilt agentd; restarting the sandbox with the current runtime picks the feature up.

## Verification

- agentd: all 115 unit tests pass in a Linux container, including a new interleave-order test (32 alternating stdout/stderr writes read back in exact emission order, zero stderr events). Protocol schema snapshot test confirms no unintended surface drift.
- Rust SDK: two new integration tests pass in a real microVM (freshly cross-compiled agentd).
- Python SDK: new integration test passes in a real microVM.
- Go / node-ts: integration cases added (compile/vet clean; VM runs covered by CI).
- CLI smoke: `--combined-output` yields `one/two/three` in order; default behavior unchanged.
- `cargo clippy`, `fmt`, and `cargo doc` (`-D warnings`) all clean.

Docs: new **Combined output** section in `docs/sandboxes/commands.mdx`, the CLI flag table, and per-language entries in each `docs/sdk/<lang>/execution.mdx` reference.

https://claude.ai/code/session_01NwJyWnd1KkshSrt4vCFC47

### DIFF
--- a/crates/agentd/lib/session.rs
+++ b/crates/agentd/lib/session.rs
@@ -267,7 +267,8 @@ impl RawActivity {
 impl ExecSession {
     /// Spawns a new exec session.
     ///
-    /// If `req.tty` is true, uses a PTY. Otherwise, uses piped stdin/stdout/stderr.
+    /// If `req.tty` is true, uses a PTY. Otherwise, uses piped stdin/stdout/stderr;
+    /// with `req.combined_output`, stderr shares the stdout pipe (`2>&1`).
     /// A background task is spawned to read output and send events via `tx`.
     pub fn spawn(
         id: u32,
@@ -546,8 +547,16 @@ impl ExecSession {
         let mut cmd = Command::new(&req.cmd);
         cmd.args(&req.args)
             .stdin(Stdio::piped())
-            .stdout(Stdio::piped())
-            .stderr(Stdio::piped());
+            .stdout(Stdio::piped());
+
+        // For combined output the child's stderr is pointed at the stdout
+        // pipe via dup2 in pre_exec (below), so no stderr pipe is created;
+        // sharing one pipe is what preserves the interleaved write order.
+        if req.combined_output {
+            cmd.stderr(Stdio::null());
+        } else {
+            cmd.stderr(Stdio::piped());
+        }
 
         for var in &req.env {
             if let Some((key, val)) = var.split_once('=') {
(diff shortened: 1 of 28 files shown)


ya-luotao:
Closing this as stale. The v0.7 agent transport changes mean it would need a rework rather than a rebase. If combined output (#723) is still wanted, let me know on the issue and I'll redo it against the current protocol.

PR #1247, opened 2026-07-30
feat(sdk): expose insecure/ca-certs registry overrides in Go and Python

### LINKED ISSUE #1184: Expose insecure/ca-certs registry overrides through the Go and Python SDK
Having an HTTPS requirement for pulling images during the `CreateSandbox()` call when `WithImage()` is passed as an option with a local Docker registry currently fails if you are not running the local registry over HTTPS.

Having this requirements means that host machines must have custom certs installed, in order to take advantage of the local Docker registry. Supplying install scripts to handle creating and installing root certs get very complicated quickly when you account for the various OSes. It also just feels off to have to do that, just to pull an image from the same machine.

My use case is for an agent harness. The harness starts external services via Docker, including a local registry on start up. Ideally, as I know the registry will always be on the same host as the harness, I would like to use HTTP for the image pull. The local registry is required, as the images used are built on the host as part of the harness installation, in order to detect and match the correct versions of applications installed on the host in the sandbox image.

It would be much nicer to be able to flag the image pull to use HTTP instead, much like we already can with the CLI tool.

ya-luotao:
Closes #1184

## What

Exposes the per-registry transport overrides that already exist in the Rust core (and the global `config.json`) through the Go and Python SDKs, so a sandbox pulling from a local plain-HTTP registry or a private-CA registry no longer needs host-level config:

**Go**
```go
sb, err := m.CreateSandbox(ctx, "worker",
    m.WithImage("localhost:5050/my-app:latest"),
    m.WithRegistryInsecure(),
    m.WithRegistryCACertsPath("/etc/ssl/private-ca.pem"), // or WithRegistryCACerts(pem)
)
```

**Python**
```python
sb = await Sandbox.create(
    "worker",
    image="localhost:5050/my-app:latest",
    registry_insecure=True,
    registry_ca_certs=["/etc/ssl/private-ca.pem", pem_bytes],  # paths or inline PEM
)
```

## Design notes

- **Single `registry()` application.** `SandboxBuilder::registry` overwrites `insecure`/`ca_certs` on every call, so auth + insecure + CA certs are merged into one closure application on both language paths.
- **Eager file reads.** CA-cert paths are read at `CreateSandbox()` / `create()` time so an unreadable file fails up front rather than mid-pull. Go option functions only record paths (they can't return errors); the read happens in `CreateSandbox`.
- **Python entry typing.** Each `registry_ca_certs` entry is either PEM data (`bytes`/`bytearray`) or a PEM file path (`str`/`os.PathLike`). Bytes are checked first — `os.fspath` also accepts bytes paths, so probing for a path first would misread inline PEM as a filename.
- API naming mirrors the existing surfaces: Rust's builder `insecure()`/`ca_certs()`, TS's `caCerts`, and the Go SDK's `WithRegistryAuth` option family.

## Verification

- Go: new unit tests for option accumulation and the FFI JSON envelope (`buildFFICreateOptions` asserted without booting the runtime); `go test` / `go vet` clean.
- Python: new `tests/test_registry_overrides.py` covering kwarg validation (bad types, unreadable paths, bytes-vs-path disambiguation) plus parse-stage acceptance; full Python SDK unit suite passes.
- Docs: `docs/sdk/go/sandbox.mdx` (new option entries), `docs/sdk/python/sandbox.mdx` (two SandboxConfig rows), and `docs/images/overview.mdx` (Python/Go examples in the registry-override section, replacing the stale Go placeholder comment that pointed at `config.json`).

https://claude.ai/code/session_01NwJyWnd1KkshSrt4vCFC47

### DIFF
--- a/docs/sdk/python/sandbox.mdx
+++ b/docs/sdk/python/sandbox.mdx
@@ -1180,7 +1180,7 @@ The keyword arguments accepted by [`create()`](#sandbox-create) and [`create_wit
 | log_level | `str \| `[`LogLevel`](#loglevel) | - | Override log verbosity |
 | registry_auth | [`RegistryAuth`](#registryauth) | - | Private registry credentials |
 | registry_insecure | `bool` | `False` | Pull images over plain HTTP instead of HTTPS (local or self-hosted registries) |
-| registry_ca_certs | `list[bytes \| str \| os.PathLike]` | `[]` | Extra CA roots to trust when pulling. Each entry is either PEM data (`bytes`/`bytearray`) or the path of a PEM file, read when `create()` is called |
+| registry_ca_certs | `list[bytes \| bytearray \| str \| os.PathLike]` | `[]` | Additional CA roots for image pulls. File paths are read when `create()` is called. |
 | volumes | `dict[str, `[`MountConfig`](/sdk/python/volumes)`]` | `{}` | Volume mounts. See [Volumes](/sdk/python/volumes) |
 | patches | `list[`[`PatchConfig`](#patchconfig)`]` | `[]` | Rootfs modifications applied before boot |
 | ports | `dict[int, int] \| Sequence[`[`PortBinding`](/sdk/python/networking#portbinding)`]` | `{}` | Port mappings. Dict form is TCP and binds to `127.0.0.1`; use `PortBinding` for explicit bind addresses or UDP |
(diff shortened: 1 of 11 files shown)


docs/sdk/python/sandbox.mdx
```diff
@@ -1179,6 +1179,8 @@ The keyword arguments accepted by [`create()`](#sandbox-create) and [`create_wit
 | pull_policy | `str \| `[`PullPolicy`](#pullpolicy) | `"if-missing"` | Image pull behavior |
 | log_level | `str \| `[`LogLevel`](#loglevel) | - | Override log verbosity |
 | registry_auth | [`RegistryAuth`](#registryauth) | - | Private registry credentials |
+| registry_insecure | `bool` | `False` | Pull images over plain HTTP instead of HTTPS (local or self-hosted registries) |
+| registry_ca_certs | `list[bytes \| str \| os.PathLike]` | `[]` | Extra CA roots to trust when pulling. Each entry is either PEM data (`bytes`/`bytearray`) or the path of a PEM file, read when `create()` is called |
```
toksdotdev:
```suggestion
| registry_ca_certs | `list[bytes \| bytearray \| str \| os.PathLike]` | `[]` | Additional CA roots for image pulls. File paths are read when `create()` is called. |
```

toksdotdev (approved):
asides that comment, lgtm!

PR #1252, opened 2026-08-01
feat(sdk/go): support AttachWith for non-default guest users

### LINKED ISSUE #1237: Support `AttachWith` in Go SDK
Via the GO SDK, I would like to attach to a running VM with a different user as the VM default user. Python and Typescript SDKs support this, the GO SDK does not.

IIITManjeet:
The Go SDK could only attach as the sandbox default user: its FFI wire struct carried only argv, and the Rust shim called sb.attach() instead of sb.attach_with(), so a user override had nowhere to go. Python and TypeScript have supported this since attach was introduced.

Extend the attach options passed across the FFI boundary to include cwd, user, env, and detach_keys, route the Rust shim through attach_with(), and add AttachWith with WithAttachUser, WithAttachCwd, WithAttachEnv, and WithAttachDetachKeys. The existing Attach API remains unchanged, making this a fully additive change.

Attach requires a real TTY and cannot be exercised reliably in automated tests, so coverage is limited to the option setters, consistent with the existing note in integration/config_test.go.

Closes #1237


IIITManjeet:
Hey @toksdotdev @appcypher can you please have a look over to my changes and if any improvements will implement those as well

toksdotdev (approved):
hey, thanks for the addition! much appreciated. i made tiny change to the docs, but asides that, lgtm! 😃

IIITManjeet:
Hey @toksdotdev somehow one of the integration test got timed out can you please once approve the tests again.
Thanks

IIITManjeet:
Hey @toksdotdev, would you mind taking a look at why this specific integration test is timing out? We've tried rerunning it three times, but it's consistently timing out. Any pointers or suggestions would be greatly appreciated. Thanks!

PR #1254, opened 2026-08-01
feat(network): report policy denials to a host-side observer

### LINKED ISSUE #691: Capture & expose network policy deny events
### Motivation

We're building a sandbox firewall management UI on top of microsandbox: users maintain an allowlist of egress hosts, the sandbox enforces it via `NetworkPolicy`, and we want to surface "denied: pkg.example.com:443" to the user so they can decide whether to add the host with one click.

The blocker is that we have no programmatic way to observe deny events from the SDK side. The policy decisions happen silently to host-side code as well, so we can't react to them.

### Current state (0.4.5)

The deny path already logs internally via `tracing::debug!` — TLS proxy, DNS forwarder, ICMP relay all have it (if I'm reading the source right). But these aren't exposed through the SDK's public surface, so host-side code can't read them.

### Ask

Could you expose deny events through some public API? We don't have a strong preference for the exact shape — whatever fits microsandbox's architecture direction (a stream method, an extension to `LogSource`, an agent-protocol message type, or something else entirely).

Per event we'd find useful: source kind (tls / dns / tcp / icmp), host or IP if available, port if available, timestamp.

nosey-dewdrop:
## TL;DR

Closes #691. Adds a `PolicyObserver` on `SharedState` so embedding code can see every network policy denial, reported from the evaluation chokepoints rather than the call sites.

## Description

- While reading the deny paths for #691 I found the reporting is uneven: the policy decision is reached from eight call sites, and only three of them emit anything (`icmp_relay.rs`, `dns/forwarder.rs`, `publisher.rs`). The TCP proxy and the TLS proxy refuse silently. Nothing is exposed to host-side code on any path.
- New `crates/network/lib/policy/observer.rs`: `PolicyDenial` with a `DenialTarget` that separates a resolved peer address from a DNS name refused before resolution, plus the `PolicyObserver` trait.
- `SharedState` gains `policy_observer` next to the existing `termination_hook`, with `set_policy_observer` and `report_policy_denial`. `SharedState` is already threaded through every evaluator, so no evaluator signature changes.
- The report sits at the chokepoints: `egress_walk` and the ingress rule walk were each split so the denial has a single exit point to observe. A new caller of the egress or ingress API cannot forget to report.
- DNS reports from `dns/forwarder.rs` instead, because `evaluate_dns_query` and `evaluate_dns_query_without_name` take no `SharedState` and widening those public signatures would be a breaking change. The reason is written next to the call.
- The denial is built lazily via `FnOnce`, so nothing is allocated when no observer is installed.
- Hostname comes from the TLS `HostnameSource::Sni` when the decision used it, otherwise from the resolved-hostname index. No new public accessor was added for that.

The event shape covers what #691 asked for, minus the timestamp: the observer stamps it, so the evaluation path does not read the clock per denied packet. Happy to add it to the struct instead if you prefer.

## Test Plan

New tests in `policy/observer.rs`, five cases: a denied egress reports destination and port, an allowed egress reports nothing, the portless ICMP path reports `None`, a denied ingress reports the guest listening port, and a denial with no observer installed is a no-op.

```
cargo test -p microsandbox-network --lib policy::observer
    5 passed; 0 failed

cargo test -p microsandbox-network --lib policy::
    124 passed; 0 failed

cargo fmt --all -- --check     clean
cargo clippy -p microsandbox-network --all-targets -- -D warnings     clean
```

Not run: the workspace-wide `cargo test --workspace` and the integration tests that need a VM.

AI usage: the implementation was written with an AI assistant. I read and own every line, and the design decisions above (chokepoint over call site, observer on `SharedState` rather than on the serde-serialized `NetworkPolicy`, DNS reported at the forwarder to avoid a breaking signature change) are mine.

### DIFF
--- a/crates/network/lib/dns/forwarder.rs
+++ b/crates/network/lib/dns/forwarder.rs
@@ -43,7 +43,7 @@ use super::common::config::NormalizedDnsConfig;
 use super::common::filter::{is_private_ipv4, is_private_ipv6};
 use super::common::transport::Transport;
 use super::nameserver::{read_host_dns_servers, resolve_nameservers};
-use crate::policy::{Action, DomainName, NetworkPolicy};
+use crate::policy::{Action, DomainName, NetworkPolicy, PolicyDenial};
 use crate::shared::{ResolvedHostnameFamily, SharedState};
 use crate::stack::GatewayIps;
 
@@ -184,6 +184,16 @@ impl DnsForwarder {
         // the DNS protocol/port.
         if decide_dns_action(&self.network_policy, &domain, transport).is_deny() {
             tracing::debug!(domain = %domain, "DNS query denied by network policy");
+            // Reported here rather than inside the policy evaluator: the
+            // DNS evaluators take no `SharedState`, and widening their
+            // public signatures would break consumers of this crate.
+            self.shared.report_policy_denial(|| {
+                PolicyDenial::to_domain(
+                    transport.policy_protocol(),
+                    domain.clone(),
+                    Some(transport.upstream_port()),
+                )
+            });
             // NXDOMAIN, not REFUSED: stub resolvers (e.g. glibc) don't
             // fail-fast on REFUSED, so a denied lookup hangs the guest in a
             // deny-by-default sandbox. NXDOMAIN is a synthetic negative that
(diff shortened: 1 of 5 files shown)


toksdotdev:
@nosey-dewdrop sorry for the delay on this, and thanks a lot for opening the pr. 

per the original thread you referenced, our intent is to approach this in quite a different, but consistent way. however, i do like your approach, and could potentially be a temporary workaround. i'll need to think a bit more about this.

PR #1257, opened 2026-08-02
chore(deps): upgrade sea-orm to 2.0 and sqlx to 0.9

connyay:
Adapt to the sea-orm 2.0 / sqlx 0.9 API changes:

- sqlx 0.9 split the combined runtime feature; use runtime-tokio + tls-rustls. Enable sqlite-use-returning-for-3_35 explicitly (a sea-orm default we opt out of via default-features = false).
- ConnectionTrait's required methods are now execute_raw / query_one_raw / query_all_raw; the old builder-taking names delegate to them, so moving the write-pool retry_on_busy wrapping onto the _raw methods preserves retry semantics for every ORM-level write.
- RuntimeErr::SqlxError now wraps Arc<sqlx::Error>; match through as_ref.
- Import ExprTrait where .eq().and() chains are built (sea-query 1.0 moved `and` onto the trait), and use UFCS for serde_json::Value::is_null where the sea-query prelude would shadow it with ExprTrait::is_null.

Add retry tests that drive a genuine SQLITE_BUSY through the sqlx driver, pinning the exact error shape the busy classifier depends on so a future driver upgrade that changes it fails the test instead of silently disabling cross-process write retries.

https://www.sea-ql.org/blog/2025-10-20-sea-orm-2.0/
https://www.sea-ql.org/blog/2026-01-12-sea-orm-2.0/

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,11 +58,24 @@ dependencies = [
 
 [[package]]
 name = "ahash"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
+dependencies = [
+ "getrandom 0.2.17",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
+name = "ahash"
 version = "0.8.12"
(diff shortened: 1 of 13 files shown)


toksdotdev (approved):
thanks a lot for this change! much appreciate 🙏🏽

PR #1258, opened 2026-08-02
feat(network): add port-scoped Rule constructors

### LINKED ISSUE #1236: Hardening: allow_egress/deny_* convenience constructors default to all-ports
### This is a hardening suggestion, not a vulnerability report

The policy engine behaves exactly as documented. This is about how easy it is to build an *unintentionally* wide rule with the convenience constructors, based on a review of a project that consumes this crate.

### Observation

`Rule::allow_egress(Destination::Group(DestinationGroup::Host))` (and the other `allow_*`/`deny_*` convenience constructors) leave `ports` empty, which means "all ports". That default is documented (`policy/types.rs:111-114`), but the intent is invisible at the call site: a reader sees `allow_egress(Host)` and does not necessarily register that it granted every port, not a specific service.

In the consuming project this produced a "host-only" posture that, correctly per this crate, allowed the guest to reach *every* host TCP port rather than just the one forward-server port it needed. The engine did the right thing; the call site expressed more than the author intended.

### Suggestions (any one of these would help)

1. A port-scoped convenience constructor, e.g. `Rule::allow_egress_ports(dest, [PortRange{..}])`, so the common "one service" case is as short as the all-ports case.
2. A doc note on the `allow_*` constructors themselves (not only on the `ports` field) stating that the result matches all ports.
3. Optionally, a `#[must_use]`-style lint or builder step that makes "all ports" a deliberate choice rather than the path of least resistance.

Happy to send a PR for (1) + (2) if that shape sounds right.

IIITManjeet:
`Rule::allow_egress(Destination::Group(DestinationGroup::Host))` and the other
`allow_*`/`deny_*` convenience constructors leave `ports` empty, which means "all
ports". The default is documented on the `ports` field, but it is invisible at the call
site: a reader sees `allow_egress(Host)` and does not necessarily register that it
granted every host port rather than the one service it needed.

Add `allow_egress_ports` / `deny_egress_ports` and the ingress and any-direction
counterparts, each taking an `IntoIterator<Item = PortRange>`, so scoping a rule to one
service is as short as the all-ports form. An empty iterator degrades to the existing
all-ports behavior — both paths funnel through the same private constructor, so the
equivalence is structural rather than two implementations that happen to agree.

Also state the all-ports default on the constructors themselves rather than only on the
`ports` field, and note there that a non-empty port set excludes ICMP (the ICMP
evaluation path skips any rule that filters on ports).

Docs: `docs/sdk/rust/networking.mdx` marks the existing constructors as all-ports and
adds a table plus an example for the port-scoped variants.

Suggestion (3) from the issue is not included: `NetworkPolicyBuilder` / `RuleBuilder`
already provide the deliberate port-scoping path via `.port()`, `.port_range()` and
`.ports()`. The gap was only in the raw `Rule::*` constructors.

No behavior change to existing rules or the wire format — the change is purely additive
and serde is untouched.

## Test plan

- `cargo test -p microsandbox-network` — 453 passed, 0 failed
- `cargo clippy -p microsandbox-network --all-targets` — clean
- `cargo fmt --check` — clean

Three tests added to `policy::types::tests`:

- `port_scoped_constructors_set_direction_action_and_ports` — each of the six sets
  direction/action/ports and leaves protocols any-protocol
- `port_scoped_constructors_with_no_ports_match_the_all_ports_constructors` — an empty
  `ports` argument yields a rule equal to the all-ports constructor, for all six pairs
- `port_scoped_host_rule_grants_only_the_named_service` — the reported scenario: under
  `default_egress = Deny`, `allow_egress(Group(Host))` permits gateway:22, while
  `allow_egress_ports(Group(Host), [8080])` permits :8080 and denies :22

I did not run the full integration suite locally (it needs a working microVM host);
happy to narrow down anything else that applies.

Closes #1236

### DIFF
--- a/crates/network/lib/policy/types.rs
+++ b/crates/network/lib/policy/types.rs
@@ -109,7 +109,8 @@ pub struct Rule {
     pub protocols: Vec<Protocol>,
 
     /// Port-range set (empty = any port). Always the guest-side port:
-    /// destination port for egress, listening port for ingress.
+    /// destination port for egress, listening port for ingress. A
+    /// non-empty set also excludes ICMP, which carries no ports.
     #[serde(default)]
     pub ports: Vec<PortRange>,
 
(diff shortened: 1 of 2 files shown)


IIITManjeet:
Hey @toksdotdev can you also review this PR as well.
Thanks :)

toksdotdev:
thanks for the pr! we decided to close this wrt https://github.com/superradcompany/microsandbox/issues/1236#issuecomment-5246303432.

PR #1269, opened 2026-08-03
feat(filesystem): allow staged agentd in offline builds

### LINKED ISSUE #1268: microsandbox-filesystem cannot use a staged agentd in offline builds
## Summary

A clean downstream build of `microsandbox-filesystem` with the default `prebuilt` feature downloads `agentd` from the corresponding GitHub release when no repository-local or cached binary is available.

Downstream consumers currently have no supported way to provide an `agentd` binary that was fetched separately. As a result, clean builds fail in offline or hermetic environments even after all Cargo-managed dependencies have been fetched.

Reproduced with `microsandbox-filesystem = "=0.6.8"`.

## Reproduction

Use the following Dockerfile:

```dockerfile
# syntax=docker/dockerfile:1

FROM rust:slim-bookworm

RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        build-essential \
        pkg-config \
        libcap-ng-dev \
    && rm -rf /var/lib/apt/lists/*

WORKDIR /repro

RUN cargo init --lib . \
    && printf '\nmicrosandbox-filesystem = "=0.6.8"\n' >> Cargo.toml

# Fetch everything managed by Cargo while networking is available.
RUN cargo fetch

# Build with Cargo offline and the complete build step network-isolated.
RUN --network=none cargo build --offline -vv
```

Run it with:

```bash
docker build --no-cache --progress=plain .
```

The final step fails after the build script attempts to download:

```text
https://github.com/superradcompany/microsandbox/releases/download/v0.6.8/agentd-x86_64
```

The Cargo dependencies are available locally, but `agentd` is obtained independently by the crate's build script.

## Expected behavior

Downstream consumers should have a supported, opt-in way to provide an existing `agentd` binary and skip the release download.

## Proposed behavior

Support an optional `MSB_AGENTD_PATH` environment variable in the default `prebuilt` path.

When selected, the supplied binary would be copied into `OUT_DIR` and the release download would be skipped. An invalid selected path would fail with a direct path-specific error, while the existing repository-local, cached, download, and non-`prebuilt` paths would remain available.

I have a focused implementation that I am currently using for a Nix-based project. PR: #1269.

georgrybski:
Closes #1268

## TL;DR

Allow downstream `microsandbox-filesystem` builds to supply a pre-fetched
`agentd` through `MSB_AGENTD_PATH`, so offline and hermetic builds do not need
to use the release-download fallback.

## Description

- Add an optional `MSB_AGENTD_PATH` override to the `prebuilt` build path.
- Preserve the selection order: repository-local `build/agentd`, `MSB_AGENTD_PATH`, cached `OUT_DIR/agentd`, then the GitHub release download.
- Read the supplied path through `std::env::var_os`.
- Register the environment variable and selected file with Cargo's rerun directives.
- Fail with a direct path-specific error when `MSB_AGENTD_PATH` is consulted but does not point to an existing file.
- Update the shared copy helper to remove an existing destination before replacement, allowing an output copied from a read-only source to be replaced on a later build.
- Keep the non-`prebuilt` artifact-selection logic unchanged.
- Document the downstream build interface in `DEVELOPMENT.md`.

## Test Plan

Repository checks:

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace -- -D warnings`
- [x] `cargo test -p microsandbox-filesystem` — 619 passed
- [x] `cargo build -p microsandbox-cli`
- [x] `pre-commit run --all-files`

Build-script behavior:

- [x] With no repository-local binary, a valid `MSB_AGENTD_PATH` is copied into `OUT_DIR`.
- [x] A network-isolated BuildKit build succeeds with the staged binary, emits no release-download log, and produces output byte-identical to the supplied file.
- [x] Repository-local `build/agentd` takes precedence over a different `MSB_AGENTD_PATH`.
- [x] An invalid `MSB_AGENTD_PATH` is not consulted when repository-local `build/agentd` is available.
- [x] With no repository-local binary, an invalid `MSB_AGENTD_PATH` fails with the direct path-specific error.
- [x] With neither a local nor explicit binary, an existing `OUT_DIR` copy is reused without downloading.
- [x] A previous output copied from a read-only source can be replaced.
- [x] Changing `MSB_AGENTD_PATH` reruns the build script through `rerun-if-env-changed`.
- [x] Changing the selected file reruns the build script through `rerun-if-changed`.
- [x] Without a local or explicit binary, the network-isolated control build attempts the existing release download and fails as expected.


toksdotdev:
@georgrybski thanks a lot of the change. much appreciate. i'm merging this now since prior build passed. another build should get triggered on main.

PR #1277, opened 2026-08-04
fix(image): preserve OCI config labels in inspection

lirazsiri:
## TL;DR

Preserve parsed OCI image labels when persisting image metadata so SDK and CLI image inspection returns the labels present in the source image config.

## Description

- serialize non-empty OCI config labels into the image config database record
- retain the existing null representation for images without labels
- add a migrated SQLite backend test covering the persistence-to-inspection round trip

Image labels commonly carry capability, provenance, lifecycle, and tooling metadata. Callers should be able to inspect that metadata through the local image API without implementing a separate registry client or downloading the config blob again.

The parser already retained labels in the in-memory image metadata. The database upsert discarded them by always writing `NULL`, so `Image::inspect` and `msb image inspect --format json` reported no labels even for labeled images.

## Test Plan

- `cargo +1.94.0 fmt --all -- --check`
- `cargo +1.94.0 test -p microsandbox image::tests`
- `cargo +1.94.0 clippy -p microsandbox -- -D warnings`


toksdotdev:
~could we also backfill labels on the [fast path](https://github.com/superradcompany/microsandbox/blob/9c3f14e93544357807a8f8a7336cdc41c0eea242/sdk/rust/lib/image/mod.rs#L1042)? if incoming labels are non-empty and stored labels are missing/different, we should reconcile. this is especially useful for cached images with `labels = null`.~

asides that, lgtm! thanks for the fix. 🥳

toksdotdev:
thought about my prior comment a bit more, and no need paying for this query on every fast path when it's temporary. anyone who needs the label on already cached image should purge that image, and refetch.

merging as is. thanks a lot 🎉

toksdotdev (approved):

PR #1286, opened 2026-08-06
Reap adopted guest children

lirazsiri:
## Summary
- reap zombie children adopted by the normal guest PID 1 agent
- preserve exit-status ownership for active exec-session leaders

## Root cause
The normal agent waited for direct exec-session children and reaped during shutdown, but it did not reap descendants adopted after their session parent exited.

## Validation
- `cargo test -p microsandbox-agentd --lib`
- real microVM probe that orphaned a Python child and verified its `/proc` entry disappeared

### DIFF
--- a/crates/agentd/lib/agent.rs
+++ b/crates/agentd/lib/agent.rs
@@ -1,6 +1,6 @@
 //! Main agent loop: serial I/O, session management, heartbeat.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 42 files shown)


appcypher:
this has been fixed in https://github.com/superradcompany/microsandbox/pull/1212

toksdotdev:
thanks for the fix, so proceeding to close this (see comment above). if the behaviour is still observed against `main`, feel free to re-open.

PR #1306, opened 2026-08-10
feat(mount): add uid=/gid= owner override for host-created files

lyb2320:
Bind and named mounts gain `uid=<N>` / `gid=<N>` options that set the guest owner presented for host files which have no per-file stat override.

Files created directly on the host (outside the guest) have no override, so they surface with the runtime's fallback owner: the guest's resolved default user on Unix, or 0:0 on Windows. When the sandbox process runs as a different user than the one owning the host files, those files become unreadable inside the guest. `uid=`/`gid=` let a caller pin the owner explicitly (e.g. `uid=1000,gid=1000`), following the vfat/ntfs/cifs/9p convention for filesystems without native ownership.

- parse_mount_spec: parse uid=/gid= (rejected unless given as a pair)
- Unix: construct the BindIdentityMap host-side from the explicit owner; the relay's default-user install now yields to an already-installed map
- Windows: present the configured owner in the no-override stat path instead of the hardcoded 0:0
- threaded through MountOptions, the MountBuilder owner() setter, the spawn mount-spec renderer, the Go FFI MountSpec, and the Go SDK MountOptions
- unspecified uid/gid preserves the prior per-platform behavior

Known limitation (Unix: shared identity map)

On Unix, the bind-identity map is a single shared handle (Arc<OnceLock<BindIdentityMap>>) registered once via BindIdentityMapRegistration and reused by every stat-virtualized mount in the VM. When a mount pins an explicit owner, bind_identity_map_for_mount installs it with handle.set(...), which is first-wins: a later mount that sets a different uid=/gid= is silently ignored (let _ = handle.set(...)).

Consequences:
- Multiple mounts using the same explicit owner (e.g. every bind mount pinned to uid=1000,gid=1000) — fine.
- A single mount with uid=/gid= — fine.
- Multiple mounts requesting different owners on Unix — only the first-registered mount's owner takes effect for all of them; the others' values are dropped without error.

This does not affect Windows, where each mount carries its own default_owner on its PassthroughConfig, so per-mount owners are honored independently.

Lifting this would mean giving each virtualized mount its own BindIdentityMap handle instead of the shared registration handle; it was left as a single shared map here to match the existing BindIdentityMapRegistration design, since the motivating use case pins the same owner on all mounts.

### DIFF
--- a/crates/cli/lib/commands/common.rs
+++ b/crates/cli/lib/commands/common.rs
@@ -128,10 +128,16 @@ pub struct SandboxOpts {
     pub volume: Vec<String>,
 
     /// Explicitly mount a host directory into the sandbox (`SOURCE:DEST[:OPTIONS]`).
+    ///
+    /// OPTIONS may include `uid=<N>,gid=<N>` to present host-created files (no
+    /// per-file stat override) as that guest owner; both are required together.
     #[arg(long = "mount-dir", value_name = "SOURCE:DEST[:OPTIONS]")]
     pub mount_dir: Vec<String>,
 
     /// Explicitly mount a host file into the sandbox (`SOURCE:DEST[:OPTIONS]`).
+    ///
+    /// OPTIONS may include `uid=<N>,gid=<N>` to present the host file (no
+    /// per-file stat override) as that guest owner; both are required together.
     #[arg(long = "mount-file", value_name = "SOURCE:DEST[:OPTIONS]")]
     pub mount_file: Vec<String>,
 
@@ -554,6 +560,8 @@ struct CliMountOptions {
     named_kind: Option<VolumeKind>,
     fstype: Option<String>,
     format: Option<DiskImageFormat>,
+    override_uid: Option<u32>,
+    override_gid: Option<u32>,
 }
 
 /// Which keyed options are valid for a public CLI mount flag.
@@ -565,6 +573,7 @@ struct CliMountOptionSupport {
     named_kind: bool,
     fstype: bool,
     format: bool,
+    owner: bool,
 }
 
 /// Parsed `SOURCE:DEST[:OPTIONS]` mount specification.
@@ -1490,6 +1499,7 @@ pub fn apply_explicit_dir_mount(
         CliMountOptionSupport {
             policies: true,
             quota: true,
+            owner: true,
             ..CliMountOptionSupport::default()
         },
     )?;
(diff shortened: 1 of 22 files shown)


lyb2320:
Supplement:  my OCI image has an special entrypoint script,  it does some work under root, then drop to normal user, so I can't use "USER  foo" in Dockerfile, then microsandbox  uses root as default user for the host files which has no xattr meta data,  that makes  injected files by host os whose perm is 0600 can't be read in vm.

lyb2320:
The patch is verified on msb CLI and microsandbox go sdk,  issues raised by greptile-apps bot are already fixed, could you take a look?  @appcypher  @toksdotdev 

```bash
$ cat > msb-entitlements.plist <<'EOF'
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
      <key>com.apple.security.hypervisor</key>
      <true/>
      <key>com.apple.security.cs.disable-library-validation</key>
      <true/>
</dict>
</plist>
EOF

$ codesign --entitlements msb-entitlements.plist --force -s - ./msb-darwin-arm64

# verify the entitlement is present
$ codesign -d --entitlements - ./msb-darwin-arm64

$ cat a.sh
#!/usr/bin/env bash
set -euo pipefail
MSB="${MSB:-msb}"
HOST="$PWD/hostdir"
mkdir -p "$HOST"
#trap 'rm -rf "$HOST"; "$MSB" rm msb-owner-before msb-owner-after >/dev/null 2>&1 || true' EXIT
echo hi > "$HOST/hello.txt"   # host-created file, owned by you, NOT uid 4242

echo "== BEFORE (no uid/gid; prior default-user behavior) =="
"$MSB" run --name msb-owner-before --replace alpine \
  --mount-dir "$HOST:/data" \
  -- stat -c '%u:%g' /data/hello.txt

echo "== AFTER (uid=4242,gid=4242) =="
"$MSB" run --name msb-owner-after --replace alpine \
  --mount-dir "$HOST:/data:uid=4242,gid=4242" \
  -- stat -c '%u:%g' /data/hello.txt

$ MSB=./msb-darwin-arm64 bash a.sh
== BEFORE (no uid/gid; prior default-user behavior) ==
0:0
== AFTER (uid=4242,gid=4242) ==
4242:4242
```

appcypher:
Superseded by #1451 so we can continue the work on a maintainer-writable branch. The three original commits from this PR are preserved unchanged in the replacement, including their authorship. Thank you @lyb2320 for the implementation.

PR #1332, opened 2026-08-13
fix(windows): preserve unicode in interactive terminals

### LINKED ISSUE #1230: Windows PTY attach corrupts UTF-8 console output
## Description

Interactive sandbox output is decoded incorrectly on Windows when the guest
writes UTF-8 characters. ASCII text and ANSI colors work, but box-drawing
characters, symbols, and spinner glyphs are rendered as mojibake.

This makes terminal applications difficult or impossible to use. For example,
Unicode borders and spinners in full-screen TUIs expand into several visible
characters, which can also throw off cursor positioning and line clearing.

## Minimal reproduction

Open PowerShell in Windows Terminal and run:

```powershell
msb --version
chcp 65001

$hex = "756e69636f64653a20e2809420e29c9320e2a08b0a"
$code = "import sys;sys.stdout.buffer.write(bytes.fromhex('$hex'))"

msb run python --tty -- python3 -c $code
```

The command contains ASCII only. Python writes this exact UTF-8 byte sequence
from inside the guest:

```text
unicode: — ✓ ⠋
```

## Expected behavior

The command should display:

```text
unicode: — ✓ ⠋
```

## Actual behavior

The console reports that code page 65001 is active, but the output is rendered
as:

```text
unicode: ΓÇö Γ£ô Γáï
```

Running the same command through the non-interactive output path works:

```powershell
msb run python --no-tty -- python3 -c $code
```

```text
unicode: — ✓ ⠋
```

This isolates the corruption to the interactive PTY attach path.

## Environment

- Microsandbox 0.6.7
- Windows 11
- Windows Terminal
- PowerShell
- Console code page 65001

## Additional context

The Windows attach implementation enables virtual-terminal input and output,
so ASCII ANSI colors and cursor controls work. It then forwards guest output
bytes directly to the console with `WriteFile`:

https://github.com/superradcompany/microsandbox/blob/b0e8eaad7a4d53fe2fbf1211a906a2d6bf3617cf/sdk/rust/lib/sandbox/attach.rs#L643-L709

The input path similarly reads console bytes with `ReadFile`:

https://github.com/superradcompany/microsandbox/blob/b0e8eaad7a4d53fe2fbf1211a906a2d6bf3617cf/sdk/rust/lib/sandbox/attach.rs#L724-L780

The Windows console documentation recommends Unicode applications use the
wide-character console APIs. A possible fix would be to incrementally decode
guest UTF-8 output and write it with `WriteConsoleW`, preserving incomplete
UTF-8 sequences between transport chunks. The input path may need the matching
`ReadConsoleW` and UTF-8 encoding treatment.

Relevant Windows documentation:

- https://learn.microsoft.com/windows/console/console-code-pages
- https://learn.microsoft.com/windows/console/high-level-console-modes
- https://learn.microsoft.com/windows/console/high-level-console-input-and-output-functions

toksdotdev:
## TL;DR

fixes windows interactive terminal unicode handling for sandbox attach and ssh. fixes #1230.

## Description

- the console path was treating utf-8 pty bytes as console bytes, which mangled non-ascii output
- use the wide windows console apis and preserve split utf-8 sequences and utf-16 surrogate pairs across chunks
- move the shared terminal handling into a private sdk module used by attach and ssh

### DIFF
--- a/sdk/rust/lib/sandbox/attach.rs
+++ b/sdk/rust/lib/sandbox/attach.rs
@@ -53,6 +53,12 @@ pub(crate) struct DetachKeys {
     sequence: Vec<u8>,
 }
 
+#[cfg(any(windows, test))]
+#[derive(Default)]
+struct WindowsUtf8Decoder {
+    pending: Vec<u8>,
+}
+
 //--------------------------------------------------------------------------------------------------
 // Methods
 //--------------------------------------------------------------------------------------------------
@@ -218,6 +224,50 @@ impl DetachKeys {
     }
 }
 
+#[cfg(any(windows, test))]
+impl WindowsUtf8Decoder {
+    fn decode(&mut self, data: &[u8]) -> Vec<u16> {
+        self.pending.extend_from_slice(data);
+
+        let mut decoded = Vec::new();
+        let mut offset = 0usize;
+        while offset < self.pending.len() {
+            match std::str::from_utf8(&self.pending[offset..]) {
+                Ok(valid) => {
+                    decoded.extend(valid.encode_utf16());
+                    offset = self.pending.len();
+                }
+                Err(error) => {
+                    let valid_end = offset + error.valid_up_to();
+                    decoded.extend(
+                        std::str::from_utf8(&self.pending[offset..valid_end])
+                            .expect("UTF-8 validation identified a valid prefix")
+                            .encode_utf16(),
+                    );
+                    offset = valid_end;
+
+                    let Some(error_len) = error.error_len() else {
+                        break;
+                    };
+                    decoded.push(char::REPLACEMENT_CHARACTER as u16);
+                    offset += error_len;
+                }
+            }
+        }
+
+        self.pending.drain(..offset);
+        decoded
+    }
+
+    fn finish(&mut self) -> Vec<u16> {
+        let decoded = String::from_utf8_lossy(&self.pending)
+            .encode_utf16()
+            .collect();
+        self.pending.clear();
+        decoded
+    }
+}
+
 //--------------------------------------------------------------------------------------------------
 // Functions
 //--------------------------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


boostd:
damn you beat me to it :D
finally got a chance to work on it but it looks like i picked the same day as you did

anyways i also got empirical confirmation of the input-side defect under `chcp 437`
```powershell
$c = "import sys; d=sys.stdin.buffer.readline(); print('RECEIVED:', d.hex())"
msb run python --tty -- python3 -c $c
```
typing é with unpatched delivers 820a (CP437), with my patch c3a90a (UTF-8)

and also for the output side guest PTY reads are 4096 bytes so this puts a multi-byte char across the frame boundary at each offset
```powershell
foreach ($pad in 4090..4096) {
  $c2 = "import sys;sys.stdout.buffer.write(b'a'*$pad + bytes.fromhex('454e44e28094e29c93e2a08b') + b'\n')"
  msb run python --tty -- python3 -c $c2
}
```

one small review note, `write_console` passes the whole buffer (chunked only at `u32::MAX`) so if `WriteConsoleW` ever returns a short write it can split a surrogate pair across calls. idk if it can ever bite given frame sizes but `std` does chunk `MAX_BUFFER_SIZE / 2` and explicitly completes an orphaned low surrogate

my branch is at https://github.com/boostd/microsandbox/tree/fix-windows-console-utf8 if anything might be useful

toksdotdev:
haha, indeed a coincidence. 😅

i'd rather you get on the contributor list, other than myself. so feel free to open a PR, and i'll close mine in favor of yours (no need to reference my pr). plus, it looks like you've also got some things that slipped the crack in mine.

i kinda prefer the new terminal module in mine, so if you could preserve it, that'd be great :)

boostd:
wonderful!
i added the refactor for a terminal module and created the PR :)

https://github.com/superradcompany/microsandbox/pull/1336

PR #1336, opened 2026-08-13
fix(windows/tty): use WriteConsoleW/ReadConsoleW for interactive attach

### LINKED ISSUE #1230: Windows PTY attach corrupts UTF-8 console output

boostd:
## TL;DR

Interactive attach on Windows forwarded guest bytes through the console's byte-oriented APIs, which
reinterpret them in the current console code page and corrupted every non-ASCII character. Both
directions now use the wide APIs (`WriteConsoleW`/`ReadConsoleW`) with incremental UTF-8/UTF-16
conversion, so guest output and typed input are correct regardless of the active code page.

Fixes #1230

## Description

`WriteFile` on a console handle behaves as `WriteConsoleA`: it decodes the bytes using the console's
current output code page before rendering. Guest output is UTF-8, so on a typical console (CP437)
`—` (`E2 80 94`) was rendered as `ΓÇö`. ASCII and ANSI escape sequences were unaffected, which is why
colors worked while box-drawing characters and spinner glyphs did not.

`ReadFile` on a console handle behaves as `ReadConsoleA` and has the mirror-image defect: typed
characters were encoded into the console's input code page instead of UTF-8. Confirmed on Windows 11
under `chcp 437` by having the guest print the bytes it received as hex:

- before: typing `é` delivered `82` (CP437)
- after: typing `é` delivers `c3a9` (UTF-8)

The wide console APIs are code-page independent, but speak UTF-16, so the host must convert. The
conversion has to be incremental: guest output arrives in frames sized by the guest's PTY reads
(4096 bytes, `crates/agentd/lib/session.rs`), so a multi-byte character routinely straddles a frame
boundary, and console input can split a surrogate pair across two reads. A code-page fix alone would
not have addressed this second defect, since each `WriteFile` call was converted independently.

The Windows console handling is first moved out of `attach.rs` into a shared
`sdk/rust/lib/sandbox/terminal/` module, since `ssh.rs` was reaching into it through
`attach::agent::` to drive its own interactive session. That first commit is pure code movement.
`terminal/encoding.rs` then holds the conversion:

- `Utf8ToUtf16Decoder` — guest bytes to UTF-16, carrying an incomplete trailing sequence into the
  next chunk, with `finish()` flushing anything still pending when the session ends. Uses
  `Utf8Error::valid_up_to()`/`error_len()` to distinguish "incomplete" from "invalid", which
  `String::from_utf8_lossy` cannot do.
- `Utf16ToUtf8Decoder` — console UTF-16 to UTF-8, carrying an unpaired high surrogate into the next
  read.
- `surrogate_safe_split` — chunks writes without splitting a surrogate pair.

This mirrors what `std`'s own Windows stdout does (`library/std/src/sys/stdio/windows.rs`: an
`incomplete_utf8` carry, `write_console_utf16`, and chunking at `MAX_BUFFER_SIZE / 2` units), which is
also why the existing `--no-tty` path was already correct.

Two notes for reviewers:

- Bytes that are not valid UTF-8 now render as `U+FFFD` rather than as code-page glyphs, and never
  raise an error, so a guest writing binary data cannot break the session. This is a deliberate
  behaviour change beyond the reported symptom.
- `SetConsoleOutputCP(CP_UTF8)` was considered and rejected: the code page is global to the console
  and shared with the user's shell, and it would leave the split-sequence defect in place. Using the
  wide APIs also means users no longer need `chcp 65001` at all.

The `terminal` module is gated `#[cfg(any(windows, test))]`. The `test` arm is deliberate: the Windows CI job
runs `cargo check`/`clippy` plus one named `microsandbox-runtime` test, so a `cfg(windows)`-only test
would never execute anywhere. Gating on `any(windows, test)` lets the conversion logic (which makes no
Win32 calls) be covered by the existing Linux/macOS test job, while unix release builds do not compile
it at all.

Because every SDK routes attach through this code via FFI, and `sandbox/ssh.rs` shares the same
`WindowsTerminalGuard`, this covers `msb run --tty`, `msb exec -it`, `msb ssh`, and the SDK attach
APIs. No dependency, `Cargo.toml`, lockfile, or CI changes: `Win32_System_Console` was already enabled.

## Test Plan

Automated:

```
cargo test -p microsandbox --lib                 # 575 passed, including 23 new encoding tests
cargo clippy --workspace --exclude microsandbox-agentd -- -D warnings
cargo fmt --all -- --check
cargo build -p microsandbox-cli
cargo clippy -p microsandbox --no-default-features --features net,ssh \
      --target x86_64-pc-windows-gnu -- -D warnings
```

The new unit tests cover the issue's exact byte sequence, a 4-byte character split at every interior
boundary, invalid and binary input, surrogate pairs split across reads, and a one-byte-at-a-time
round trip.

Manual, on Windows 11 / Windows Terminal / PowerShell:

1. Issue repro under `chcp 65001` renders `unicode: — ✓ ⠋`.
2. Same repro under `chcp 437` also renders correctly, confirming code-page independence.
3. Frame-boundary sweep: for pad in 4090..4096, the guest writes `b'a'*pad` followed by
   `END—✓⠋`, placing a multi-byte character across the 4096-byte PTY read boundary. All offsets
   render correctly.
4. Guest writing 4096 random bytes produces replacement characters, exits cleanly, and leaves the
   terminal usable.
5. Input, under `chcp 437`, with the guest printing the bytes it received as hex: `é` arrives as
   `c3a9`, `日` as `e697a5`, and `🚀` as `f09f9a80` — correct UTF-8 at all three sequence widths,
   the last of which is a surrogate pair on the console side. The unpatched build delivered `82`
   for `é`.
6. No regressions: ANSI colors, `Ctrl+]` detach, and window resize propagation all behave as before.

```mermaid
flowchart LR
    GuestOut[Guest UTF-8 output] --> U8Decoder[Incremental UTF-8 to UTF-16 decoder]
    U8Decoder --> SafeSplit[Surrogate-safe chunks]
    SafeSplit --> WriteW[WriteConsoleW]
    ReadW[ReadConsoleW] --> U16Decoder[Incremental UTF-16 to UTF-8 decoder]
    U16Decoder --> Detach[Detach-sequence detection]
    Detach --> GuestIn[Guest UTF-8 input]
    Attach[Direct attach] --> Terminal[Shared Windows terminal module]
    SSH[Interactive SSH] --> Terminal
    Terminal --> U8Decoder
    Terminal --> ReadW
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;main&#39; into fix-windows-con..."](https://github.com/superradcompany/microsandbox/commit/44b1710c426ab331c0bc199143ee23aa5ced3b5b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53418197)</sub>


toksdotdev:
this looks good! one small thing: in `attach.rs`, could we propagate the errors from `terminal_guard.write_output()` and `finish_output()` instead of using `let _ =`? the ssh path already uses `?`, so this would keep both paths consistent and avoid silently dropping console write failures.

boostd:
> this looks good! one small thing: in `attach.rs`, could we propagate the errors from `terminal_guard.write_output()` and `finish_output()` instead of using `let _ =`? the ssh path already uses `?`, so this would keep both paths consistent and avoid silently dropping console write failures.

done 👍
just in case also retested on windows, no regressions.

toksdotdev (approved):
lgtm! thanks a lot for making the fix. much appreciated 🙌🏾

PR #1373, opened 2026-08-17
fix(sandbox): contain bind rootfs patches

appcypher:
## TL;DR

Contain bind-rootfs patch operations inside the configured guest root, including through descendant symlinks and Windows junctions. Harden replacement semantics so `replace: false` is atomic, preserve guest-internal symlink behavior, and retain the existing `follow_root_symlinks` opt-in for selecting a linked host root.

## Description

- Open protected bind roots with a handle-relative, component-by-component no-follow walk and keep the resulting directory capability pinned for the patch operation.
- Resolve existing symlinks with Linux guest-root semantics, so absolute targets restart at the guest root and parent traversal clamps at that root.
- Resolve every recursive `CopyDir` destination independently, including nested host-absolute symlinks.
- Preflight complete bind-patch batches before mutation, including Windows-incompatible modes, symlinks, and host-path syntax.
- Use atomic no-clobber creation for `replace: false`, apply modes through pinned handles, and use one-shot POSIX creation modes for new directories.
- Perform text, file, copy, directory, symlink, remove, append, and permission mutations through `cap-std` instead of ambient host paths.
- Add native Windows junction coverage and run focused bind-patch tests in both Windows CI matrix jobs.
- Document that nested mounts and hard-link aliases are selected host capabilities outside path-based containment.

## Test Plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo check -p microsandbox --locked`
- [x] `cargo clippy -p microsandbox --lib -- -D warnings`
- [x] `cargo test -p microsandbox --lib sandbox::patch` (26 passed)
- [x] macOS ARM64 live Alpine/libkrun scenarios for all patch operations, nested `CopyDir`, final symlinks, and both replacement modes
- [x] Windows ARM64 focused native security suite before the final race-hardening pass (8 passed), including descendant junction escapes and linked-root policy
- [x] Fresh Windows ARM64/x64 CI for the final mode-handle, canonical-path, and reserved-name cases
- [x] Windows ARM64 full SDK suite exercised (589 passed, 1 ignored; 2 unrelated failures in runtime CLI-argument and remote named-pipe/datagram tests)
- [x] macOS full SDK suite exercised (one unrelated runtime-probe test was flaky in the full run and passed immediately in isolation)

```mermaid
flowchart TD
    A[Bind rootfs patch batch] --> B[Validate complete batch]
    B --> C[Open and pin configured bind root]
    C --> D[Resolve guest path components]
    D --> E{Existing symlink?}
    E -- Absolute target --> F[Restart at guest root]
    E -- Relative target --> G[Continue from current directory]
    F --> D
    G --> D
    E -- No --> H[Capability-relative mutation]
    H --> I[Atomic create or pinned-handle replacement]
    I --> J[Apply mode through opened handle]
    J --> K[Proceed to next patch]
```
</details>

<sub>Reviews (7): Last reviewed commit: ["ci: bound apt installs on hosted runners"](https://github.com/superradcompany/microsandbox/commit/85b151a4703347cbd407682d613e45f34997ae8e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53902587)</sub>


appcypher:
## Live rootfs containment validation

Validated commit `f1146be5` on macOS/aarch64 with a freshly built and codesigned `msb`, current `agentd`, and a bootable Alpine 3.22 bind root. Each successful case booted a real libkrun VM and verified the result from inside the guest; host-side sentinels were checked after shutdown.

- [x] All bind patch operations through a host-absolute descendant symlink: Text, File, CopyFile, CopyDir, Mkdir, Append, Remove, and Symlink remained under the guest root. The external host directory and its sentinel contents were unchanged.
- [x] Relative parent symlink (`../outside`) clamped at the guest root. The external sibling was unchanged.
- [x] Legitimate internal absolute symlink (`/compat -> /etc`) remained compatible and guest-visible.
- [x] Final symlink semantics: Append followed the target relative to the guest root; Remove unlinked the symlink itself. Both external host targets were unchanged.
- [x] Final-component bind-root symlink was rejected by default and accepted only with `follow_root_symlinks(true)`.
- [x] Intermediate-component bind-root symlink was rejected by default and accepted only with `follow_root_symlinks(true)`.
- [x] Full GitHub CI is green, including macOS/Linux tests and Windows builds.

### Remaining completeness gaps

- Windows CI compiles this path but does not execute the SDK bind-patch tests. Junction/reparse-point behavior still needs Windows-native tests before claiming equal runtime validation across all supported hosts.
- Unsupported Windows mode and symlink patches currently perform some filesystem work before returning their platform error. This behavior predates this PR, but should be changed to validate before mutation.
- `CopyDir` now remains contained when it encounters a nested destination symlink, but an internal absolute symlink below the destination may be rejected rather than resolved with guest-root semantics. This is a compatibility edge, not an observed escape.
- `replace=false` still has a check-then-create race. Capability-relative operations preserve containment, but atomic no-replace semantics should be handled separately.
- Nested mounts and pre-existing hard-link aliases are outside path-based capability containment and need an explicit threat-model decision.

Conclusion: the reported descendant-symlink host escape is contained in the native scenarios tested, but the broader cross-platform patch subsystem should not yet be described as exhaustively complete.

appcypher:
Follow-up hardening is now pushed in verified signed commit `fe4649c5`. This resolves the three validation gaps from the prior test note: atomic `replace: false`, nested `CopyDir` symlink containment, and native Windows junction coverage.\n\nNative results:\n- macOS ARM64: 18 focused unit tests pass plus live Alpine/libkrun escape scenarios across all patch operations.\n- Windows ARM64: 8 focused security tests pass, including descendant junction escapes, linked-root policy, batch preflight, and concurrent no-clobber creation.\n- Windows full SDK: 589 passed, 1 ignored, with 2 unrelated failures in tests outside the changed patch code.\n\nThe Windows CI matrix now runs the focused bind-patch suite on both ARM64 and x64.

appcypher:
Final adversarial review found and fixed two additional TOCTOU classes in verified signed commit `cce240fa`:\n\n- file modes are applied through the already-open file handle;\n- mode-bearing directories are prepared privately, chmodded through a directory handle, and published with OS-level atomic no-replace rename on Linux/macOS;\n- explicit `..` destination components are rejected before mutation, avoiding symlink-dependent semantic changes within a batch;\n- Windows preflight now rejects reserved device names and the complete invalid filename-character set.\n\nThe focused macOS suite is now 25/25, with deterministic rename-substitution tests for both files and directories. Fresh Windows matrix jobs are pending; the physical ARM64 host's SSH service began timing out during banner exchange after the prior successful 8/8 native run, so the PR test plan records that distinction explicitly.

appcypher:
Superseding the temporary-directory detail in my previous note: final verified signed commit `b311763a` removes temporary publication entirely. New mode-bearing directories now use one atomic POSIX `mkdir(mode)` operation, so a concurrent writer cannot substitute the object between creation and chmod; the host umask may only narrow those creation bits. Existing directories are opened first and receive the exact mode through their pinned handle, including when the parent is not writable. Focused result: 26/26.

appcypher:
Final Windows gate is green on commit `d312dc79`: both `Windows Quality (windows-aarch64)` and `Windows Quality (windows-x86_64)` passed check, clippy, bind-root backend tests, and the focused bind-root patch suite. The PR test plan is updated accordingly.

PR #1435, opened 2026-08-22
feat(sdk): resolve Go SDK install dir from MSB_HOME

lyb2320:
## What

Resolve the Go SDK's `installDir()` from `$MSB_HOME` when set, falling back to `~/.microsandbox` otherwise. This is the root under which the SDK materializes the embedded FFI library (`lib/v<ver>/`) and, in download mode, fetches and looks up `msb` + `libkrunfw` (`bin/`, `lib/`).

## Why

The runtime's `resolve_home()` already honors `$MSB_HOME` for the state db, sandboxes, and cache, but `installDir()` was hardcoded to `~/.microsandbox`. So `MSB_HOME` could relocate the runtime state, yet the Go SDK's FFI/runtime files always went to the global home.

That split means an instance's on-disk footprint ends up half in a private root and half in the shared global home, and there is no single env that fully isolates a Go SDK consumer's microsandbox files (e.g. to run isolated instances side by side, or to avoid one binary's runtime clobbering another's). Resolving both sides from the same env removes the inconsistency.

## How

```go
func installDir() (string, error) {
	if h := os.Getenv("MSB_HOME"); h != "" {
		return h, nil
	}
	// ~/.microsandbox fallback (unchanged)
}
```

- Matches `resolve_home()`'s use of `MSB_HOME`, so the FFI/runtime root and the state root agree.
- The per-version FFI subdir (`lib/v<ver>/`) is unchanged.

## Compatibility

No behavior change when `MSB_HOME` is unset: consumers still share `~/.microsandbox` so `msb` + `libkrunfw` aren't redownloaded per binary. Setting `MSB_HOME` now contains the whole footprint under one root.

## Test

Added `TestInstallDir_HonorsMSBHome` covering both the `MSB_HOME`-set and unset (default `~/.microsandbox`) paths.


toksdotdev (approved):
hey, thanks a lot for the fix! much appreciated. 🙏🏾

PR #1447, opened 2026-08-23
feat(mounts): support guest uid and gid overrides

appcypher:
## TL;DR
Add paired per-mount `uid`/`gid` options that present virtiofs-backed host files with a fixed guest-visible owner without changing host ownership.

## Description
- Accept paired `uid=<id>,gid=<id>` options for bind mounts and directory-backed named volumes in the CLI and YAML configuration.
- Preserve per-inode `user.msb.override_stat` metadata as the highest-precedence source; the mount identity is only the fallback for paths without an override.
- Use a dedicated fixed identity map on Unix and equivalent ADS/in-memory fallback behavior on Windows.
- Expose the options through the Rust, Node.js, Python, and Go SDKs, including validation, generated types, inspect output, and documentation.
- Reject partial pairs, non-virtiofs mounts, directory-incompatible named disk volumes, and `stat-virt=off`, which promises literal host metadata.
- Omit unset ownership fields from serialized mount options so existing wire payloads remain unchanged.

The new optional fields extend the public Rust `MountOptions` struct. Builder-based callers are unaffected; callers using exhaustive struct literals may need to add `..Default::default()`.

## Test Plan
- [x] `cargo fmt --all --check`
- [x] `cargo test -p microsandbox-filesystem -p microsandbox-runtime -p microsandbox -p microsandbox-cli -p microsandbox-types --features microsandbox-types/ts`
- [x] `cargo clippy -p microsandbox-filesystem -p microsandbox-runtime -p microsandbox -p microsandbox-cli --lib -- -D warnings`
- [x] `go test ./...` in `sdk/go`
- [x] `pytest tests/test_mount_policies.py`, `ruff check`, and `ruff format --check` in `sdk/python`
- [x] `npm run typecheck` and `npm run test:unit` against a freshly rebuilt native addon in `sdk/node-ts` (130 tests)
- [x] `git diff --check`
- [ ] Run the native Windows filesystem suite in CI; the current macOS host cannot compile the Windows-native dependencies without the MSVC SDK.

### DIFF
--- a/crates/cli/lib/commands/common.rs
+++ b/crates/cli/lib/commands/common.rs
@@ -549,6 +549,8 @@ struct CliMountOptions {
     follow_root_symlinks: bool,
     stat_virtualization: Option<microsandbox::sandbox::StatVirtualization>,
     host_permissions: Option<microsandbox::sandbox::HostPermissions>,
+    uid: Option<u32>,
+    gid: Option<u32>,
     size_mib: Option<u32>,
     quota_mib: Option<u32>,
     named_kind: Option<VolumeKind>,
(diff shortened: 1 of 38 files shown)


appcypher:
Superseded by #1451, which preserves the original contributor's commits and incorporates the reviewed ownership fixes.

PR #1448, opened 2026-08-23
feat(filesystem): add bind-mount deny lists to hide host files/paths

### LINKED ISSUE #1326: Option to hide/exclude paths inside a host bind mount from the guest
# Motivation

I'm working on a sandbox for opencode using microsandbox. It launches opencode inside a microsandbox VM, bind-mounting
the host project directory as `/workspace`. The agent legitimately needs full read-write access to the working tree to
edit code, but the goal is to keep credentials out of the sandbox so they cannot be read or exfiltrated by a
compromised, careless, or malicious agent.

I'm already providing options for using microsandbox secrets, but many projects keep secrets in conventional on-disk
files — `.env`, `.envrc`, and framework-specific derivations such as `.env.local`, `.env.production`, or my tool's own
`env.secret.yaml` — which are simply passed through by the bind mount. Even with read tool deny patterns configured in
opencode (`"permission":{"read": {"*": "allow","*.env": "deny",...}}`, an agent can trivially bypass them with a shell one-liner (`cat .env`).

A host-controlled mechanism to hide those paths from the guest would guarantee that the agent can edit code freely while not being able to read the files that hold credentials.

# Proposal

Add a mount option that hides a set of paths from the guest for a bind mount:

- Hidden paths are invisible: 
    - `readdir`/`readdirplus` omit matching entries.
    - `ENOENT` on `lookup`/`open`(read)/`getattr`.
- Writes to hidden paths are forbidden: `EACCES` on `create`/`mkdir`/`mknod`/`symlink`/`link`/`unlink`/`rmdir`/`rename`/`open`(write).
- Enforcement is on the host, in the filesystem backend.
- The rest of the tree stays fully read-write.

## Deny-pattern syntax

I assume excellent UX by following the [gitignore pattern format](https://git-scm.com/docs/gitignore#_pattern_format). For simplicity, an initial design could implement a subset. 

## Example Usage

```rust
microsandbox.Mount.Bind(hostDir, microsandbox.MountOptions{
    Deny: []string{".env", ".env.*", ".envrc"},
})
```

```text
--mount-dir /abs/host:/workspace:deny=.env,deny=.env.*,deny=.envrc
```

neurolabs:
Closes #1326

## Summary

Adds an optional deny list of gitignore-style patterns to bind mounts. Matching host paths are hidden from the guest (ENOENT) and writes are rejected (EACCES) via the Unix/Windows passthrough backends. Usable across the CLI and Rust/Go/Python/Node SDKs.

## Description

- Shared `DenyList` matcher in `crates/filesystem`
- Enforcement at the lookup gate plus create/mkdir/mknod/symlink/link/unlink/rmdir/rename, with readdir entry filtering, on both the unix and Windows passthrough backends (rename re-checks the destination so hidden entry types don't leak)
- basename patterns, path patterns, and dir-only patterns (`node_modules/`)
- per-mount case-sensitivity: probed similar to git's `core.ignorecase` for rw mounts, derived from platform/filesystem type on ro mounts
- macOS resolves parent paths via `/.vol` + `F_GETPATH` against the canonical mount root and fails closed when a path cannot be resolved
- performance varies with cases and patterns (tested on linux:
  - lookup overhead is negligible, denies are faster than actual lookups.
  - readdir overhead is substantial:
    - basename only (e.g. "*.log") and dirname patterns (e.g. "node_modules/") cost about 1.2x compared to a baseline
    - path patterns (e.g. "/var/log/*.log") cost about <2 compared to a baseline
  - see crates/filesystem/lib/backends/passthroughfs/unix/tests/test_deny_perf.rs for a test suite
- Wire format: repeatable `deny=` mount option; runtime and SDKs all reject empty patterns and patterns containing `,`/`:`/newline/NUL.
- CLI: `deny=` on `--mount-dir` only (rejected with a clear error on `--mount-file`/`--mount-disk`/`--mount-named`); `msb inspect` renders a `[deny=...]` suffix
- Cloud: deny lists rejected at the `CloudCreateBody` conversion boundary (cloud creates can't enforce them)
- Docs (volumes, CLI reference, security/filesystem) document the semantics and the name-based limitations: rename evasion on writable mounts, symlink targets not re-checked, hard links matched by path not inode

## Notes

- Developed and tested on Linux, therefore the MacOS and Windows paths were neither compiled nor executed.

<details><summary><h3>Confidence Score: 3/5</h3></summary>

The PR does not yet appear safe to merge because the outstanding Unix exchange race can still place a directory at a directory-only-denied path.

The reverse-direction exchange check uses a destination type captured before the host mutation, but only the source identity is revalidated; an external replacement of the destination can therefore invalidate the policy decision before the atomic exchange.

**Files Needing Attention:** crates/filesystem/lib/backends/passthroughfs/unix/remove_ops.rs
</details>

<sub>Reviews (15): Last reviewed commit: ["chore(filesystem): some review fixes"](https://github.com/superradcompany/microsandbox/commit/b581dee6277960cb598580137c0a61fa8d3d476e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56024017)</sub>

**Context used:**

- Knowledge Base — [Sandbox filesystem architecture](https://app.greptile.com/microsandbox/-/custom-context/knowledge-base/superradcompany/microsandbox/-/docs/filesystem.md)
- Knowledge Base — [Filesystem backends and operations](https://app.greptile.com/microsandbox/-/custom-context/knowledge-base/superradcompany/microsandbox/-/docs/filesystem-backends.md)


crates/filesystem/lib/backends/passthroughfs/windows/ops.rs
```diff
@@ -152,22 +167,106 @@ impl DynFileSystem for PassthroughFs {
...
+            // When the filesystem reports no file identity (FAT32/exFAT, some
+            // network volumes) the source's type at move time cannot be verified.
+            // On such filesystems this guard makes a rename fail closed (EBUSY)
+            // only when the destination collides with a dir-only pattern and the
+            // source was decided to be a file, because that is the one case a swap
+            // could turn into a directory landing at a denied name. Ordinary
+            // renames between non-denied paths, and renames of a same-named file
+            // away from a dir-only pattern, are unaffected: FAT/exFAT keeps nearly
+            // full rename support except for the narrow destination-collision case.
+            let after_identity = file_identity(&self.safe_metadata(&old_path)?);
```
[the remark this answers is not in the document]
neurolabs: I won't fix that. Rationale:

* This is a deliberately accepted residual risk, already through 4 review rounds.
* The identical residual window exists in the Linux and macOS backends. Neither kernel offers a rename-conditional-on-inode syscall, so it cannot be closed on those platforms. Eliminating it only on Windows changes the project's overall exposure negligibly while creating cross-backend asymmetry.
* Guest requests are serialized by the single-threaded virtio-fs worker. Exploitation needs a hostile co-located host actor threading a sub-millisecond window, and the payoff is content landing at a path the victim mount thereafter hides (ENOENT on lookup).
* The only complete Windows fix is disproportionate and risky: open the source handle without FILE_SHARE_DELETE and rename via SetFileInformationByHandle(FileRenameInfo). That means large unsafe-FFI changes with real behavior changes: concurrent host processes (AV, indexers, other msb mounts) would start getting ERROR_SHARING_VIOLATION (→ guest-visible EBUSY) on legit deletes/renames of the source; new failure modes requiring DELETE access on the source; differing semantics vs MoveFileEx across NTFS/FAT/exFAT/SMB.
* I can't validate the Windows platform, don't use it.

I will extend comments, though.

crates/filesystem/lib/backends/passthroughfs/unix/tests/test_deny.rs
```diff
@@ -0,0 +1,765 @@
...
+/// moves a denied path-pattern match to a non-denied name.
+///
+/// `deny: ["sub/.env"]` is location-based. Renaming `sub` -> `x` succeeds
+/// (neither `sub` nor `x` matches the pattern), and the subsequent lookup of
+/// `x/.env` reconstructs the path from the new anchor, so the file is served.
+/// This is the documented confidentiality limitation of path patterns on a
+/// writable mount (see docs/sandboxes/volumes.mdx); it is intentionally
+/// `#[ignore]`d because the bypass is current behavior, not a bug to fail on.
+#[test]
+#[ignore = "known limitation: path patterns are location-based and laundered by ancestor rename"]
```
appcypher, line 706: i dont think we can treat this as a known limitation. if the guest can rename `sub` to `x` and then read `x/.env`, it has bypassed the deny rule. for this pr, we should either fix that or reject path patterns on writable mounts.

crates/filesystem/lib/backends/passthroughfs/deny.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// case-sensitivity is instead inferred from the filesystem type; see
+    /// [`mount_is_case_insensitive`].
+    pub(crate) fn new(root: &Path, patterns: &[String], readonly: bool) -> Self {
+        let mut builder = GitignoreBuilder::new(Path::new("/"));
+        let mut needs_path_reconstruction = false;
+        let mut has_dir_only_patterns = false;
+        for pattern in patterns {
+            needs_path_reconstruction |= pattern.trim_end_matches('/').contains('/');
+            has_dir_only_patterns |= pattern.ends_with('/');
+            let _ = builder.add_line(None, pattern);
```
appcypher: can we return an error here instead of ignoring it? if one pattern is invalid, the sandbox still starts with less protection than the user asked for. we should only start when every pattern has been accepted.

sdk/rust/lib/runtime/spawn.rs
```diff
@@ -2210,6 +2211,12 @@ fn push_dir_mount_arg(
     if let Some(mib) = quota_mib {
         opts.push(format!("quota={mib}"));
     }
+    for p in deny {
+        // Trim the pattern so a value like `" .env"` renders as `deny=.env`
+        // rather than a literal-space pattern `deny= .env`, matching the CLI
+        // and runtime, which trim option tokens on parse.
+        opts.push(format!("deny={}", p.trim()));
```
appcypher, line 2221: can we avoid trimming the pattern here? whitespace can change the gitignore meaning. for example, `" !secret"` becomes `"!secret"` and turns into an allow rule. we should preserve the value exactly or reject it with a clear error.

sdk/rust/lib/runtime/spawn.rs
```diff
@@ -2239,6 +2247,12 @@ fn push_file_mount_arg(
         options.override_gid,
     );
     opts.push(format!("quota={quota_mib}"));
+    for p in deny {
+        // Trim the pattern so a value like `" .env"` renders as `deny=.env`
+        // rather than a literal-space pattern `deny= .env`, matching the CLI
+        // and runtime, which trim option tokens on parse.
+        opts.push(format!("deny={}", p.trim()));
```
appcypher, line 2257: i think we need to reject this for file binds. we pass the deny rule here, but `SingleFileFs` does not apply it, so the sandbox starts and the file is still visible. rejecting it seems clearer than accepting a rule we dont enforce.

packages/microsandbox-types/rust/lib/domain.rs
```diff
@@ -1724,6 +1731,7 @@ impl Serialize for VolumeMount {
                 map.serialize_entry("host_permissions", host_permissions)?;
                 map.serialize_entry("follow_root_symlinks", follow_root_symlinks)?;
                 map.serialize_entry("quota_mib", quota_mib)?;
+                map.serialize_entry("deny", deny)?;
```
appcypher, line 1781: the deny list is saved in the sandbox `config` and `active_config` records. an older msb version can read those records but will ignore the unknown `deny` field. so this can happen: create a sandbox with deny rules -> downgrade msb -> restart the sandbox -> the files are exposed. can we add a compatibility-marker migration like `m20260824_000001_mount_owner_config` and have its downgrade check both records for non-empty deny lists? that would stop the downgrade with a clear error instead of dropping the policy.

crates/filesystem/lib/backends/passthroughfs/deny.rs
```diff
@@ -0,0 +1,462 @@
...
+
+/// Detect whether the filesystem holding `root` respectively the folder 'root'
+/// is case-insensitive.
+///
+/// Almost all Linux/Unix filesystems: sensitive.
+/// APFS / HFS+: insensitive by default, can be formatted case-sensitive.
+/// FAT: insensitive.
+/// NTFS: insensitive by default, can be configured per directory.
+///
+/// We don't detect case sensitivity per directory, but per mount.
```
appcypher, line 188: i dont think checking case sensitivity once at the mount root is enough. NTFS can behave differently in a child directory, so `.ENV` may still resolve to a denied `.env`. can we either check the actual parent directory or use case-insensitive matching here?

crates/filesystem/lib/backends/passthroughfs/unix/inode.rs
```diff
@@ -220,7 +216,20 @@ pub(crate) fn vol_path(dev: u64, ino: u64) -> std::ffi::CString {
 pub(crate) fn do_lookup(fs: &PassthroughFs, parent: u64, name: &CStr) -> io::Result<Entry> {
     crate::backends::shared::name_validation::validate_name(name)?;
 
+    // Determine the child's type (without following a symlink) to support
+    // dir-only deny patterns.
     let parent_fd = get_inode_fd(fs, parent)?;
+    let is_dir = if fs.deny.has_dir_only_patterns() {
+        platform::fstatat_nofollow(parent_fd.raw(), name)
+            .map(|st| platform::mode_file_type(st.st_mode) == platform::MODE_DIR)
+            .unwrap_or(false)
```
appcypher, line 225: can we avoid `unwrap_or(false)` here? if this lookup fails, we treat the entry as a file and may skip a directory-only rule. for a security check, an unknown result should return an error or stay hidden.

crates/filesystem/lib/backends/passthroughfs/unix/dir_ops.rs
```diff
@@ -402,11 +450,15 @@ fn read_dir_entries(fd: i32) -> io::Result<Vec<PassthroughDirEntry>> {
         let name =
             unsafe { std::slice::from_raw_parts(entry.d_name.as_ptr() as *const u8, name_len) };
 
+        // When the filesystem did not supply a dirent type, resolve it via
+        // `fstatat` to support dir-only deny patterns.
+        let file_type = resolve_dirent_type(libc::dirfd(dirp), entry.d_type as u32, name);
```
appcypher, line 455: this currently fails to compile on macos because `libc::dirfd` needs an unsafe block. can we fix this and add macos and windows compile checks? this PR touches both backends, so we should catch platform-specific issues before merge.

packages/microsandbox-types/rust/lib/domain.rs
```diff
@@ -397,6 +397,12 @@ pub enum VolumeMount {
         /// contents. `None` applies the protective default at spawn time; set a
         /// value to override it.
         quota_mib: Option<u32>,
+        /// Host-side deny-list of gitignore-style patterns.
+        ///
+        /// Matching entries are hidden (lookup/getattr/readdir return ENOENT) and
+        /// writes to them are forbidden (EACCES). Enforced by the host filesystem
+        /// backend; patterns are relative to the mount root.
+        deny: Vec<String>,
```
toksdotdev, line 405: once the earlier macOS compile error is fixed, `cargo test -p microsandbox --lib --no-default-features --no-run` still doesn’t compile. five existing `VolumeMount::Bind` test fixtures are missing the new required `deny` field: one each in `backend/cloud/sandbox.rs`, `backend/local/sandbox/create.rs`, and `sandbox/config.rs`, plus two in `sandbox/types.rs`. can we add `deny: Vec::new()` to those fixtures?

sdk/rust/lib/sandbox/types.rs
```diff
@@ -1206,6 +1227,34 @@ fn validate_host_path_wire_safe(path: &Path, label: &str) -> crate::Microsandbox
...
+///
+/// Patterns are rendered into the `--mount` option block as `deny=<pattern>`.
+/// The block is comma-joined and the spec is colon-split, so any of those
+/// separators (or a newline/NUL, which would corrupt gitignore parsing) must
+/// be rejected up front rather than silently attenuating other mount options.
+///
+/// Keep this in sync with the identical validators in
+/// `crates/runtime/lib/vm.rs` (`validate_deny_pattern`) and
+/// `sdk/go/native/src/lib.rs`.
+fn validate_deny_pattern(pattern: &str) -> crate::MicrosandboxResult<()> {
```
toksdotdev, line 1244: this validation currently only happens through `MountBuilder::build()`. direct or deserialized `VolumeMount::Bind` values go through `validate_volume_mount()`, but its bind arm never validates the `deny` entries. i confirmed that `deny: ["secret,ro"]` passes validation and is then serialized into the comma-delimited mount options, where `ro` becomes a real option. can we validate every bind mount’s deny list there too?

crates/filesystem/lib/backends/passthroughfs/deny.rs
```diff
@@ -0,0 +1,462 @@
...
+/// that name resolves to the same file. If it does, the filesystem folds case
+/// and deny patterns must be matched case-insensitively. The probe requires a
+/// write, so on read-only mounts (`readonly` set) it cannot run; there,
+/// case-sensitivity is instead inferred from the filesystem type ([`ro_fs_is_case_sensitive`]).
+///
+/// A failed probe defaults to `true` (case-insensitive) rather than `false`:
+/// over-matching only hides a few differently-cased names that almost never
+/// coexist, whereas under-matching on a case-insensitive host would let a
+/// pattern like `.env` be bypassed by requesting `.ENV`.
+fn mount_is_case_insensitive(root: &Path, readonly: bool) -> bool {
```
toksdotdev, line 201: even after fixing when `case_insensitive()` is configured, read-only mounts still don’t follow the actual filesystem setting here. case-sensitive APFS is always treated as case-insensitive, and Linux overlay, NFS, and CIFS mounts are too. that collapses distinct names like `.env` and `.ENV`, despite the docs promising exact matching on case-sensitive filesystems. can we detect the actual mount behavior here, or narrow the documented guarantee?

crates/filesystem/lib/backends/passthroughfs/deny.rs
```diff
@@ -0,0 +1,462 @@
...
+
+//--------------------------------------------------------------------------------------------------
+// Tests
+//--------------------------------------------------------------------------------------------------
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn deny(patterns: &[&str]) -> DenyList {
```
toksdotdev, line 322: these tests all share the same process-id temp directory while running in parallel, and each call removes it out from under the others. the full filesystem suite reproduced this as an `AlreadyExists` panic in `interior_slash_pattern_needs_path_reconstruction`. can we use a fresh `tempfile::tempdir()` for each call?

sdk/python/microsandbox/types.py
```diff
@@ -697,6 +699,7 @@ class MountConfig:
     noexec: bool = False
     nosuid: bool = False
     nodev: bool = False
+    deny: list[str] | None = None
```
toksdotdev: since `MountConfig` is a public dataclass and isn’t keyword-only, inserting `deny` here shifts the positional meaning of `disk`, `format`, `fstype`, and every field after them. an existing positional disk mount can now put its disk path into `deny` and fail serialization. could we append the new field after the existing fields so positional construction remains compatible?

crates/filesystem/lib/backends/passthroughfs/deny.rs
neurolabs: I am hesitant to write a test file to every directory on NTFS, so I lean towards using case-insensitive matching for NTFS generally. WDYT WRT test writes & deletes per directory?

crates/filesystem/lib/backends/passthroughfs/deny.rs
neurolabs: I was initially aiming for maximum possible compatibility between linux and macOS. If a macOS user uses case-sensitive APFS, deny lists should work like on linux as closely as possible.
I guess the easiest path forward is narrowing the guarantee WRT RO mounts.

crates/filesystem/lib/backends/passthroughfs/unix/dir_ops.rs
neurolabs: I would appreciate if someone using macOS could handle this. For me (on linux) it's a blind night flight ;)

crates/filesystem/lib/backends/passthroughfs/unix/tests/test_deny.rs
neurolabs: My initial thought was that the user writing deny patterns should know from documentation that this behavior is a security issue and, when handling such cases, use a non-path pattern.
This should be weighed against completely losing the possibilities of path-pattern based fine-grained denies.

crates/filesystem/lib/backends/passthroughfs/deny.rs
neurolabs: handled

crates/filesystem/lib/backends/passthroughfs/deny.rs
neurolabs: handled

sdk/python/microsandbox/types.py
neurolabs: handled

packages/microsandbox-types/rust/lib/domain.rs
neurolabs: handled (blindly)

sdk/rust/lib/sandbox/types.rs
neurolabs: handled

PR #1194, opened 2026-07-21
Add --gpu: virtio-gpu Venus (Vulkan) device for sandboxes

maxandrews:
Closes the ask in #291, structured as the separate-PRs approach you preferred — this is the flag plumbing; the `msb_krun_*` mapping fixes it depends on are **superradcompany/libkrun#91**.

**What it does:** `msb run --gpu` / `msb create --gpu` expose a virtio-gpu **Venus** device to the guest. A guest with mesa's venus userspace enumerates the host GPU and runs real Vulkan workloads.

**Plumbing:** `SandboxRuntimeOptions.gpu` (serde-default, so existing configs are untouched) → `SandboxBuilder.gpu()` → CLI flag → spawn argv → `VmConfig` → `msb_krun` ConsoleBuilder GPU flags; enables the `msb_krun/gpu` cargo feature.

**Renderer flag choices** (each earned by a debugging session):
- `VENUS | NO_VIRGL` — Vulkan encoder only, no legacy GL.
- `THREAD_SYNC | USE_ASYNC_FENCE_CB` — **required**: the device worker never calls `virgl_renderer_poll`, so synchronous fences never retire and the first real GPU submission livelocks (host pinned ~120%, looks exactly like a slow shader compile).
- Linux adds `RENDER_SERVER` — virglrenderer ≥ 1.x serves venus exclusively through the render-server proxy; without it the venus capset (id 4) is advertised but fills as **zeros**, and guests silently fall back to llvmpipe.
- `gpu_shm_size(8 GiB)` — an address-space window for venus blob mappings, not a RAM reservation.

**Also:** fixes `[linux] build-agentd` to target the host arch (was hardcoded `x86_64-unknown-linux-musl`; arm64 Linux hosts couldn't build agentd — hit while bringing this up on Graviton metal).

**Verified end-to-end:**
| Platform | Host GPU | Result |
|---|---|---|
| macOS 26 / HVF, Apple M4 Pro | MoltenVK | guest `vulkaninfo` = `Virtio-GPU Venus (Apple M4 Pro)`; llama.cpp ggml-vulkan generates tokens fully offloaded |
| Linux/KVM aarch64 (AWS g5g.metal) | NVIDIA T4G | Qwen2.5-7B Q4_K_M fully offloaded: **pp512 647.85 t/s, tg128 29.18 t/s** |
[...]

PR #1482, opened 2026-08-30
Experimental virtio-gpu display: msb display / msb run --display (macOS)

ya-luotao:
Draft: shows the whole path end to end, but it depends on three `msb_krun` changes that are not published yet (superradcompany/libkrun#116, #117, #118), vendored here under `third_party/` with `[patch.crates-io]`. Once those land and 0.1.33 is out the `third_party/` commit goes away. It also overlaps with #1194's `--gpu` plumbing: this branch gates the device with an `MSB_GPU` environment variable for now and should be rebased onto that option (`SandboxRuntimeOptions.gpu`) rather than keep its own switch.

## What it does

`msb run --display …` (or `msb display <sandbox>` on a running one) opens the guest's virtio-gpu scanout in a native macOS window with keyboard and pointer. Demo: Omarchy (Hyprland + quickshell) on Arch Linux ARM in a microVM — https://github.com/ya-luotao/msb-omarchy.

## How

- **Runtime** (`crates/runtime/lib/gpu_display/`): a `krun_display` backend copies each presented frame into a 2-slot memory-mapped file per scanout under the sandbox runtime dir and announces `configure` / `frame{slot,seq,rect}` / `disable` as JSON lines on `display.sock` (next to `agent.sock`); the viewer's evdev-style input comes back on the same socket and is queued into two virtio-input devices (keyboard; absolute pointer shaped like QEMU's tablet). `present_frame` never waits for a viewer (200 ms write timeout, then the viewer is dropped), so a stalled window cannot stall the guest's FLUSH.
- **Viewer** (`crates/cli/lib/commands/display.rs`, macOS-only deps `winit` + `softbuffer`): runs the event loop on the main thread before Tokio starts, like `msb sandbox`; BGRX frames are a straight copy into softbuffer's 0RGB buffer. The sandbox process cannot own the window because `Vm::enter()` takes its main thread.
- `MSB_GPU_DUMP=<dir>` installs a frame-dump backend instead (debugging without a window); `MSB_GPU_DISPLAY=WxH` sizes the scanout (default 1920x1080).
- Sandbox cleanup (`ipc.rs`) also removes `display.sock`.

## Findings worth knowing

- On macOS there is no virgl; the device must run on rutabaga's 2D component or every 2D command is rejected (libkrun#117).
- `msb_krun`'s `input` feature never delivered an event on macOS because of an eventfd flag bug (libkrun#116).
- The guest must **not** set `LIBGL_ALWAYS_SOFTWARE=1`: aquamarine then finds no DRM EGL device. Mesa 26 falls back to llvmpipe on its own (kms_swrast for the compositor, swrast for clients).

## Not done / open questions

- Wiring via `MSB_GPU` env instead of #1194's option; no Linux viewer (the runtime side is portable, the viewer is macOS-only by `cfg`).
- HiDPI: the window is the scanout's physical size (960x540 points on Retina); no clipboard.
- Tests: `cargo test -p microsandbox-runtime` passes; the display path itself has no unit tests yet.

https://claude.ai/code/session_01A2DMienRTQjgR5CF3uQ2AZ


ya-luotao:
Closing this experiment for now. It is too large to review as one PR and has drifted well behind main. I'm keeping the branches, and can come back with a smaller, staged proposal if there is interest in a display path.

PR #1506, opened 2026-09-02
fix(build): detect native Linux musl target

kingfs:
## Summary

The Linux `just build-agentd` recipe hard-coded `x86_64-unknown-linux-musl`. On ARM64 Linux hosts such as Kunpeng 920, Cargo invoked the native AArch64 compiler with `-m64`, causing linking to fail.

Derive the musl target from Rust's host triple so native Linux builds select the matching architecture without a fixed allowlist. Unknown host triples fail with a clear diagnostic.

## Test plan

- `just --dry-run build-agentd`
- `just build-agentd` on `aarch64-unknown-linux-gnu` (passes)
- `git diff --check`

The local commit could not be GPG-signed because the configured signing key is unavailable in this environment.

<h2><a href="https://app.greptile.com/api/retrigger?id=59324249"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; the prior native-musl build blockers are resolved and no new actionable failure was established.

<sub>Reviews (5) · Last reviewed commit: ["Merge branch &#39;main&#39; into fix/portable-ag..."](https://github.com/superradcompany/microsandbox/commit/179c34ec617d4a3065c2a44d867cecd02b548834)</sub>

### DIFF
--- a/justfile
+++ b/justfile
@@ -87,11 +87,29 @@ build-deps: build-agentd build-libkrunfw
 # Build agentd as a static Linux/musl binary. Requires: musl-tools (apt) or musl-dev (apk).
 [linux]
 build-agentd:
-    @command -v musl-gcc >/dev/null || { echo "error: musl-gcc not found. Install your distro's musl toolchain."; exit 1; }
-    rustup target add x86_64-unknown-linux-musl 2>/dev/null || true
-    cargo build --release --manifest-path crates/agentd/Cargo.toml --target-dir target --target x86_64-unknown-linux-musl
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    command -v musl-gcc >/dev/null || {
+        echo "error: musl-gcc not found. Install your distro's musl toolchain."
+        exit 1
+    }
+
+    # Derive the musl target from Rust's host triple instead of maintaining an
+    # architecture allowlist. This keeps local builds aligned with every
+    # Linux target supported by the installed Rust toolchain (x86_64, ARM64,
+    # riscv64, and future targets), while still using the native compiler.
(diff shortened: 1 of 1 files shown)


kingfs:
Greptile's finding is valid: native musl Rust hosts (for example Alpine) were rejected because the GNU-to-musl substitution left the host unchanged. Updated the recipe to preserve *-unknown-linux-musl hosts while continuing to derive musl targets from GNU hosts. Verified with just --dry-run build-agentd and git diff --check; pushed as commit 7e7280f6.

kingfs:
Evaluated Greptile's second finding: native Alpine builds were blocked by the unconditional musl-gcc check. This is valid for native musl environments. Updated the recipe so native musl hosts use their system compiler without requiring musl-gcc; GNU hosts still require musl-gcc for the musl target. Pushed commit 0dda715b after just --dry-run build-agentd and git diff --check.

toksdotdev:
@kingfs there's still an open comment. when you've got some time, could you resolve? ty!

toksdotdev:
one more comment: `rustc` installed through alpine’s package manager reports host triples such as `aarch64-alpine-linux-musl`, but the current case only accepts `*-unknown-linux-musl`, so it exits before reaching the build. could you also handle alpine’s native musl host triples here as well? thanks!

toksdotdev (approved):
i addressed the rest of the issues, and approved. thanks for the fix.

PR #1234, opened 2026-07-28
feat(network): add outbound SOCKS4 and SOCKS5 proxies

darfire:
This PR adds the option to pass all TCP connections originating from the guest apps through an external SOCKS proxy, such as mitmproxy. By setting up the correct SSL CA certificates in the sandbox the connection content can be logged and analyzed.

PR contains:
* SOCKS connection forwarding
* support for Go, Python and Typescript SDKs
* documentation and example of usage with mitmproxy

PR #1508, opened 2026-09-02
feat(network): add configurable guest proxy

m42e:
Add a configurable guest-facing HTTP forward proxy that routes sandbox HTTP and HTTPS traffic through an operator-specified upstream proxy.

- Add network.proxy.enabled and network.proxy.port configuration with CLI YAML and SDK support.
- Expose HTTP absolute-form requests and HTTPS CONNECT host:port on the sandbox gateway alias.
- Preserve destination hostnames in upstream CONNECT requests so the upstream proxy performs DNS resolution.
- Inject proxy environment variables into guests while excluding the gateway alias and loopback addresses from proxying.
- Keep direct gateway routing for published host services and retain existing host-side TCP proxy support.
- Propagate the configuration through shared domain/cloud types and the Rust, Python, Node, and Go SDKs.
- Document YAML configuration and proxy behavior.

<details><summary><h3>Confidence Score: 0/5</h3></summary>

The PR is not safe to merge while hostname proxying can bypass address-level egress restrictions and explicit proxy ports still collide across concurrent sandboxes.

Hostname targets are authorized without validating the address ultimately selected by local or upstream DNS, allowing denied private, metadata, CIDR, or destination-group endpoints to remain reachable; separately, explicit proxy ports still bind a shared loopback endpoint and prevent a second sandbox from initializing.

**Files Needing Attention:** crates/network/lib/tcp/http.rs, crates/network/lib/policy/types.rs, crates/network/lib/network.rs
</details>

<sub>Reviews (7): Last reviewed commit: ["fix(network): trust upstream DNS for all..."](https://github.com/superradcompany/microsandbox/commit/98c92de0a13a152871e92c4aa39ef8bfdb0cdb89) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59393853)</sub>

**Context used:**

- Knowledge Base — [Sandbox networking](https://app.greptile.com/microsandbox/-/custom-context/knowledge-base/superradcompany/microsandbox/-/docs/networking.md)
- Knowledge Base — [Network policy and DNS routing](https://app.greptile.com/microsandbox/-/custom-context/knowledge-base/superradcompany/microsandbox/-/docs/network-policy-and-dns.md)


toksdotdev:
hey @m42e, 

this is very timely given the concept of outbound proxy was recently added in #1234 and #1507 respectively (yet to be released).

can you adapt your changes to fit into the proxy interface that was recently introduced. it'd require adding a `OutboundProxy::HttpConnect` variant, and the interface looking similar to this:

```rust
.proxy(|p| p.socks5(address))
.proxy(|p| p.http_connect(address))
```

let me know if you have any questions. thanks.

m42e:
@toksdotdev I will look into it.

toksdotdev:
awesome! thank you.

m42e:
@toksdotdev I am not sure if my current approach is perfectly matching what you suggested.
As SOCKS does likely support more that just TCP traffic. Also I am not sure if it would work and would be easy to understand for the user if there only TCP is forwarded via the proxy. 

So currently the approach would have an http proxy server for each vm which forwards to the upstream http proxy. 

What is you opinion on that?

m42e:
An alternative approach would be to do something like the TLS interception also on PORT 80, but in some contexts also the DNS resolution is done within the proxy. So while it seems more consistent on a configuration level it might still break in some environments.

m42e:
@toksdotdev would you mind answer my question so I can update it and you could merge it?

toksdotdev:
hey @m42e, sorry for the slow reply. 

i’d still prefer to build this into the existing outbound proxy interface. `HTTP_PROXY` / `HTTPS_PROXY` can be ignored or overridden inside the sandbox, so they wouldn’t enforce the routing we want. TCP-only support is fine, provided unsupported traffic is blocked from bypassing the proxy.

your DNS point makes sense. we can address that separately by routing our DNS client’s connections through the configured outbound proxy. for HTTP CONNECT, that would mean DNS over TCP through the tunnel, assuming the proxy allows it. we could support this with SOCKS5 too.

let’s keep this PR focused on HTTP CONNECT support and handle the DNS work in a follow-up. that approach would still need a DNS server that can resolve the destination names, so let me know if your setup specifically requires the HTTP proxy itself to resolve them.

PR #1606, opened 2026-09-17
fix(network): give published-port listeners a backlog deeper than 128

yashau:
## Problem

`tcp_listener_task` binds published ports with `TcpListener::bind`, which leaves the listen backlog to mio — **128** on Linux ([mio 1.2.0](https://github.com/tokio-rs/mio/blob/v1.2.0/src/net/tcp/listener.rs), `bind` passes 128 on everything but Horizon/Haiku).

One browser can exhaust that on its own. A single page load of a modern web app makes roughly a hundred parallel requests, and a reverse proxy in front of a published port turns each into its own upstream connection. Past the queue the kernel refuses the connection, the proxy reads EOF, and a file that is present — and serves fine on retry — reaches the reader as a **502**.

That is how I found it: a Vite-built app published from a sandbox behind Caddy, returning `Failed to fetch dynamically imported module` for a chunk that was on disk.

## Measurement

Against a dev server published on a loopback port, sandbox otherwise idle:

| concurrent connections | result |
|---|---|
| 120 | 120 succeeded |
| 130 | 118 refused, 12 succeeded |
| 140 | 140 refused |

A cliff rather than a slope, which is what a fixed-length accept queue looks like — not resource exhaustion. At the time the host had load 0.20 with 18 GB free, and the forwarder was using 716 of its 524288 permitted file descriptors. `ss -lnt` reported `Send-Q 128` on the socket, matching mio's constant.

## Change

`bind_listener` builds the listener through `TcpSocket` so the backlog can be stated explicitly (1024), keeping `set_reuseaddr(true)` because `TcpListener::bind` already set it. The kernel still clamps to `net.core.somaxconn`, so a host that wants a shallower queue can still say so.

The accept loop drains quickly, so the depth only has to absorb a burst rather than sustained load.

## Test

`published_listener_queues_more_than_one_browser_can_open` opens 300 connections without accepting any, so it fails against a 128-deep queue and passes against this one.

## Test plan / what I could not run

- `cargo fmt -p microsandbox-network -- --check` — clean.
- **I could not build or run the tests locally.** My host is Windows and the workspace needs a Unix toolchain (`nix`, `portable-atomic` and `winapi` build scripts all fail at `link.exe`). The change is small and the test is self-contained, but it has not been executed — please let CI have it, and I am happy to adjust if anything fails.

<h2><a href="https://app.greptile.com/api/retrigger?id=65521238"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; both previous findings are resolved and no new actionable issue was introduced by the latest change.

<sub>Reviews (3) · Last reviewed commit: ["refactor(network): move LISTEN\_BACKLOG u..."](https://github.com/superradcompany/microsandbox/commit/01368bd521cfef2e353e3d064011fcf8bcf4ec2b)</sub>


crates/network/lib/engine/ports/publisher.rs
```diff
@@ -923,6 +955,32 @@ mod tests {
...
+        let addr = listener.local_addr().unwrap();
+
+        // Deliberately never accept: what is under test is the queue, not the accept loop.
+        let mut held = Vec::new();
+        for _ in 0..300 {
+            match tokio::time::timeout(Duration::from_secs(2), TcpStream::connect(addr)).await {
+                Ok(Ok(stream)) => held.push(stream),
+                Ok(Err(e)) => panic!("refused after {} connections: {e}", held.len()),
+                Err(_) => panic!("timed out after {} connections", held.len()),
+            }
```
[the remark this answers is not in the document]
yashau: Valid, and thank you — I checked and you are right: `check-platform.yml` runs `unit-test` → `test-platform.yml` → `cargo test --workspace` on the `macos-14` leg of the matrix, so this would have gone red there.

macOS clamps `listen()` to `kern.ipc.somaxconn`, which is 128 by default, so the 300 connections would be refused against a perfectly correct `listen(1024)` — the test would be measuring the sysctl rather than the change.

Fixed in ae79b543 by gating the test to Linux, where `net.core.somaxconn` is 4096 by default and leaves room for it. I chose that over raising the limit on the runner because configuring the host to make an assertion pass tests the configuration, not the code.

I also documented the clamp on the constant itself, since it matters to anyone publishing a port from macOS: the deeper backlog only applies as far as the host's own ceiling allows, and raising `kern.ipc.somaxconn` is what makes it count there.

crates/network/lib/engine/ports/publisher.rs
```diff
@@ -547,6 +547,40 @@ fn reject_with_rst(stream: &TcpStream) {
...
+///
+/// Measured against a published dev server on a loopback port: 120 concurrent connections all
+/// succeed, 130 leave 118 refused, 140 refuse every one. The accept loop below drains the queue
+/// quickly, so the depth only has to absorb a burst rather than sustained load.
+///
+/// The kernel clamps this to the host's own ceiling, so asking for more than it allows is not an
+/// error and a host that wants a shallower queue can still say so: `net.core.somaxconn` is 4096
+/// by default on Linux, while `kern.ipc.somaxconn` on macOS is 128 -- there, raising the sysctl
+/// is what makes this take effect.
+const LISTEN_BACKLOG: u32 = 1024;
```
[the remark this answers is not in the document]
yashau: Correct — AGENTS.md line 214 says constants and statics go under `Constants`, and this file has that section at line 31. I had declared it beside the function that uses it. Moved in 01368bd5.

crates/network/lib/engine/ports/publisher.rs
```diff
@@ -54,6 +54,24 @@ const UDP_EPHEMERAL_PORT_START: u16 = 49152;
...
```
toksdotdev, line 73: could we make this configurable through the sdk/cli, with 1024 as the default?

crates/network/lib/engine/ports/publisher.rs
```diff
@@ -54,6 +54,24 @@ const UDP_EPHEMERAL_PORT_START: u16 = 49152;
 const UDP_EPHEMERAL_PORT_COUNT: usize =
     (u16::MAX as usize) - (UDP_EPHEMERAL_PORT_START as usize) + 1;
 
+/// Backlog for a published port's listener.
+///
+/// `TcpListener::bind` leaves this to mio, which passes 128. That is not enough for a published
+/// HTTP server: one browser opening one page of a modern web app makes ~100 parallel requests, and
+/// a reverse proxy in front turns each into its own upstream connection. Beyond the queue the
+/// kernel refuses the connection, the proxy reads EOF, and a file that is present and serves fine
+/// on retry comes back as 502.
```
toksdotdev, line 63: could we describe backlog exhaustion as a possible cause here and in the pr description?

i tested locally and found 502s from the separate 256-connection cap too. i also found that already-closed connections can still count toward that cap until cleanup runs. reclaiming those slots fixed failures in repeated smaller bursts, without changing the backlog.

those can be handled separately, but it means the reported 502s alone don’t confirm backlog exhaustion.

crates/network/lib/engine/ports/publisher.rs
```diff
@@ -923,6 +957,39 @@ mod tests {
...
+    #[cfg(target_os = "linux")]
+    #[tokio::test]
+    async fn published_listener_queues_more_than_one_browser_can_open() {
+        let addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0);
+        let listener = bind_listener(addr).unwrap();
+        let addr = listener.local_addr().unwrap();
+
+        // Deliberately never accept: what is under test is the queue, not the accept loop.
+        let mut held = Vec::new();
+        for _ in 0..300 {
```
toksdotdev, line 982: this still fails on linux when `somaxconn` is 128, even though the code correctly requests 1024. i reproduced a timeout after 129 connections. could we check the host limit before expecting 300 connections to succeed?

